Issue #25, opened 2012-11-19
No way to copy the backtrace out of the messagehandler dialog


krf:
Example: http://wstaw.org/m/2012/11/19/plasma-desktopCf1097.png

There's no way to copy the backtrace from that dialog.

A good solution would be to add a "Copy backtrace" button in the dialog which copies it to the clipboard. (Just like DrKonqi does it).

krf:
Fixed by: https://github.com/KDAB/GammaRay/pull/110

PR #110, opened 2015-02-07
Add "Copy backtrace" to message dialog

agabor:
implementation of issue #25

I have seen that this issue have been abandoned for quite a while, so I took the courage and implemented it. Please review my changes!

### DIFF
--- a/ui/tools/messagehandler/messagehandlerwidget.cpp
+++ b/ui/tools/messagehandler/messagehandlerwidget.cpp
@@ -34,6 +34,10 @@
 #include <QLabel>
 #include <QListWidget>
 #include <QTime>
+#include <QPushButton>
+#include <QClipboard>
+#include <QApplication>
+#include <QSignalMapper>
 
 using namespace GammaRay;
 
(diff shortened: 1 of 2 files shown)


krf:
Code looks good. Did you test it?

Unfortunately I'm not sure *how* to test it. GammaRay changed quite a bit, and currently the message dialog is covered by the "Lost connection to remote host: The remote host closed the connection" dialog as soon as a `qFatal(...)` was registered...

agabor:
Yes I have also realized that, and I wasn't sure if thats intentional. But yes I have tested it, open the messages tab first, then cause a qFatal in the host and then the dialog will appear.

krf:
Merging this, code looks fine to me. 

We need to make sure the dialog isn't covered by the other dialog, though. Which is a different issue. Maybe you'd like look into that as well?

Thanks for the patch!

agabor:
Seems like `MessageHandlerWidget` is only created whe the "Messages" tab is first opened, otherwise the `fatalMessageReceived` signal does not get handled. I will look into this. (In two weeks at least, I am having a hollyday.)

Thank you for the feedback!

PR #112, opened 2015-03-08
Fixing some message handling issues

agabor:
This pull request is a follow up for this one: https://github.com/KDAB/GammaRay/pull/110
1. The problem mentioned there is that the message handling dialog does not always appear when a `qFatal` message is received. The reason for this is, that the `MessageHandlerWidget`, which implements the handling of this message, is lazy initialized. Because of this the message handling dialog only appears if the "Messages" tab was previously selected in the tool selector. I could think of two ways around this:
   * Force the initialization of the `MessageHandlerWidget`. This would certainly solve the problem, but feels dirty: for later readers of the code it would not be appearent why this is the only widget that is not lazy initialized.
   * Move the error handling code to somewhere else. This seemed to be the cleaner way to go, so I did this. I extracted thi implementation fo the message handling dialog into its own class, and moved the invocation code into the `MainWindow`. This also has the side effect that I can set the main window as the parent for the message handling dialog, which this way receives the GammaRay icon, improving user experience :)

2. I have run into the following phenomenon: When I start GammaRay the "Objects" tab is selected on the tool selector. Now if I click on the "Messages" tab the selector quickly jumps back to the "Objects" tab. I have found that the reason for this is how the `selectInitialTool` function is called in the `MainWindow` constructor. This function is connected to the `rowsInserted` and the `dataChanged` events of the `toolSelector`, but uppon succesfull call of this method, it disconnects itself. The problem I belevie is that this function was called from the constructor before it was connected to these events, so if the first call was succesfull it still remained connected. I assumed this was not intended, so I have chanded the invocation order, which fixed the problem.

3. A code comment in `MessageHandlerWidget` told me that the display of the backtrace for each message is a TODO, so I implemented it. For this I created the `core/tools/messagehandler/BacktraceModel` class. An instance of this class is retrieved from the `ObjectBroker` in the `MessageHandlerWidget`. I added the `selectMessage` function to the `MessageHandlerInterface` which refreshes the backtrace model when a message is selected on the GUI.


agabor:
Hello,
It has been a while. Can I have some feedback on this, please?

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+using namespace std;
+using namespace GammaRay;
+
+BacktraceModel::BacktraceModel(QObject *parent)
+{
+}
+
+int BacktraceModel::rowCount(const QModelIndex &parent) const
+{
+  Q_UNUSED(parent);
```
milianw: try to hook a ModelTester up to this model and it should spot some issues. Here for example, you should return 0 when parent is valid to indicate that this is not a tree model.

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+  return m_data.count();
+}
+
+int BacktraceModel::columnCount(const QModelIndex &parent) const
+{
+  Q_UNUSED(parent);
+  return COLUMN_COUNT;
```
milianw: similar as above, return 0 if parent.isValid()

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+//Parses input in "aaa (bbb)" format into a tuple of "aaa" and "bbb".
+tuple<QString, QString> parseParenthesis(QString str)
```
milianw: const& the str arg and wrap this helper function in an anonymous namespace to give it internal linkage.

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+  QStringList parts = str.split(" (");
+  if (parts.count() != 2)
+    return tuple<QString, QString>();
+  QString second = parts.at(1).left(parts.at(1).count()-1);
```
milianw: please add spaces around operators

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+  return make_tuple(parts.at(0), second);
+}
+
+//The input either contains a function name only,
+//or is in the ClassName::FunctionName format.
+tuple<QString, QString> parseClassAndFunctionName(QString str)
```
milianw: see above

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+        return tr("Class");
+      case FunctionColumn:
+        return tr("Function");
+    }
+  }
+
+  return QVariant();
+}
+
+void BacktraceModel::setBacktrace(Backtrace &backtrace)
```
milianw: const missing

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+  m_data.clear();
+  foreach (QString stackFrame, backtrace) {
```
milianw: const& the stackFrame

core/tools/messagehandler/backtracemodel.cpp
```diff
@@ -0,0 +1,131 @@
...
+//or is in the ClassName::FunctionName format.
+tuple<QString, QString> parseClassAndFunctionName(QString str)
+{
+  QStringList parts = str.split("::");
+  if (parts.count() == 2)
+    return make_tuple(parts.at(0), parts.at(1));
+  return make_tuple(QString(), str);
+}
+
+QStringList BacktraceModel::parseStackFrame(QString &stackFrame) const
```
milianw: const missing for the stackFrame arg

core/tools/messagehandler/messagemodel.h
```diff
@@ -57,7 +57,7 @@ class MessageModel : public QAbstractTableModel
     virtual QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const;
     virtual QVariant headerData(int section, Qt::Orientation orientation,
                                 int role = Qt::DisplayRole) const;
-
+    Backtrace getBacktrace(int idx);
```
milianw: please remove the "get" prefix, just name it "backtrace" following the Qt API naming style.

ui/mainwindow.cpp
```diff
@@ -111,9 +121,9 @@ MainWindow::MainWindow(QWidget *parent): QMainWindow(parent), ui(new Ui::MainWin
 
   setWindowTitle(tr("GammaRay (%1)").arg(Endpoint::instance()->label()));
 
-  selectInitialTool();
```
milianw: the change here means that upon selecting an initial tool, more tools get activated and that changes something? hmm that should be investigated.

milianw:
Overall this looks pretty good already, I think with some minor cleanup and then testing it this should be good to get in. many thanks!

ui/mainwindow.cpp
agabor: No, that is not what I meant, and it is actually a simpler problem.
If I understand correctly the initial tool is supposed to be selected exactly once. There are two scenarios:
 1. The `ui->toolSelector->model()` is fully populated in the constructor. I this case the initial tool is selected by the constructor.
 2. There is no fitting tool in the model. I this case the initial tool is selected, when first a fitting tool is inserted into the model. Thats why the `rowsInserted` and the `dataChanged` signal is connected to the `selectInitialTool` slot.

But if it is so, these signals should not remain connected after the `selectInitialTool` has run succesfully in the constructor, thats why I suspect that the order of these lines are intended to be as I wrote them.

Would be much cleaner, if these signal would only be connected if and when it is deteced that there are no tool to select initially.

agabor:
Hi,
Sorry for the slow processing. I fixed what you asked for, and some other bugs too. I think it should be good by now.

agabor:
Should I have done many smaller pull requests instead of this one big?

PR #117, opened 2015-03-27
Small touches

bog-dan-ro:


vkrause:
We can't do this unconditionally, as that breaks the stand-alone client. This needs to be either conditional, or moved to a place where it's only called by the plugin or the stand-alone client respectively.

bog-dan-ro:
Ah, I thought the client library is use only externally and not by the stand-alone client.

bog-dan-ro:
I added paramaters

PR #122, opened 2015-04-28
QPointer mainwindow

milianw:
only the QPointer patch is interesting, the other patch is handled in a separate review request.


vkrause:
looks good to me, feel free to merge this

Issue #109, opened 2015-02-05
Selecting "GammaRay::OverlayWidget" in Widgets tree freezes GammaRay


danvratil:
I attached GammaRay to a running process and started inspecting widgets in the "Widgets" tree. I accidentally selected a "GammaRay::OverlayWidget" item in the tree, which GammaRay injected into the widget hierarchy of the inspected application. This caused the inspected application to freeze and use 100% CPU.

Using GammaRay 2.2.1 with Qt 5.

iamsergio:
Is there any use in showing these ? Maybe they should be hidden

danvratil:
I think they are hidden by default (but I couldn't find the code), except for when a new dialog/window is created at runtime - at least that's my impression from the behaviour: I don't see the GammaRay::OverlayWidget anywhere in the tree when I start the app from GammaRay (or when I attach GammaRay to a running process), but I can see them when I open a dialog and start inspecting it - in this particular case it was KDevelop, and "Assign shortcut" dialog in Documentation toolview.

milianw:
reproduced with Qt4 and kwrite. Didn't even have to open any other dialog. 

(gdb) bt
#0  0x00007fa11f8b8a5b in QWidget::window() const () from /usr/lib/libQtGui.so.4
#1  0x00007fa11fa992ae in QWidgetPrivate::invalidateBuffer(QRect const&) () from /usr/lib/libQtGui.so.4
#2  0x00007fa11f90f5d9 in QWidgetPrivate::setParent_sys(QWidget*, QFlags<Qt::WindowType>) () from /usr/lib/libQtGui.so.4
#3  0x00007fa11f8cd2c7 in QWidget::setParent(QWidget*, QFlags<Qt::WindowType>) () from /usr/lib/libQtGui.so.4
#4  0x00007fa11f8cdfc2 in QWidget::setParent(QWidget*) () from /usr/lib/libQtGui.so.4
#5  0x00007fa08d824178 in GammaRay::OverlayWidget::placeOn(QWidget*) ()
   from /ssd/milian/projects/compiled/kde4/lib/gammaray/2.3/qt4.8-x86_64/gammaray_widgetinspector_plugin.so
#6  0x00007fa08d81fc93 in GammaRay::WidgetInspectorServer::widgetSelected(QItemSelection const&) ()
   from /ssd/milian/projects/compiled/kde4/lib/gammaray/2.3/qt4.8-x86_64/gammaray_widgetinspector_plugin.so
[...]

PR #123, opened 2015-04-30
Filter OverlayWidget and do not show it to users of GammaRay.

milianw:
This fixes https://github.com/KDAB/GammaRay/issues/109


vkrause:
Conceptually this looks good, I'm just wondering if using ProbeGuard would give the same results here, with using already existing infrastructure?

milianw:
Doh, I knew I was missing something fundamental... I was really suprised this did not exist yet - well, of course it does! I'll rework this patch to reuse the probe guard.

vkrause:
Obsolete by cedf07c and 048b384 I think.

PR #131, opened 2015-06-04
Play safe

bog-dan-ro:


vkrause:
Looks ok to me, feel free to integrate that.

PR #139, opened 2015-07-01
Make launcher library sharable and GAMMARAY_INSTALL_QT_LAYOUT

bog-dan-ro:


common/paths.cpp
```diff
@@ -51,7 +51,7 @@ void setRootPath(const QString& rootPath)
   Q_ASSERT(QDir(rootPath).exists());
   Q_ASSERT(QDir(rootPath).isAbsolute());
 
-  s_rootPath = rootPath;
+  s_rootPath = QFileInfo(rootPath).absoluteFilePath();
```
vkrause: For which case is this needed? The assert in the line above suggests rootPath is already absolute here, no?

config-gammaray.h.cmake
```diff
@@ -3,9 +3,18 @@
 // relative install dirs
 #ifdef Q_OS_ANDROID
 # define GAMMARAY_PLUGIN_INSTALL_DIR "lib"
+#elif defined(GAMMARAY_INSTALL_QT_LAYOUT)
+# define GAMMARAY_PLUGIN_INSTALL_DIR "${PROBE_PLUGIN_INSTALL_DIR}"
```
vkrause: Since this is in the elseif branch of the Android branch above, this will never be hit on Android, is that intended?

bog-dan-ro: Yes.

common/paths.cpp
bog-dan-ro: I compile it in release mode, I think I found the problem on Android ...

vkrause:
This got integrated by now, in smaller commits.

PR #140, opened 2015-07-01
Add loop detection to object tree filter.

akreuzkamp:
This commit adds a loop detection for Probe::filterObject, that will
be enabled after 100 iterations up the tree. It will then track all
visited objects in a set and check for each node, if it is already
known.


krf:
:+1:

vkrause:
Looks good to me, impact on the common case should be minimal.

PR #141, opened 2015-07-08
Add workaround for batch visualization and Qt 5.5.

akreuzkamp:
Work around a performance optimization done in Qt 5.5, that breaks
batch visualization completely. Because it breaks batch visualization,
QSGBatchRenderer will only apply it if no batch visualization is
applied. This workaround forces the batch renderer to reevaluate that,
so that it recognizes it now is applied.


krf:
Maybe target 2.2 branch?

vkrause:
No need for that I think, this is the last blocker for the 2.3.0 release.

Feel free to integrate this.

PR #143, opened 2015-07-23
Make the message handler dialog always appear when a qFatal message is received.

agabor:
This pull request is a follow up for this one: https://github.com/KDAB/GammaRay/pull/110
1. The problem mentioned there is that the message handling dialog does not always appear when a `qFatal` message is received. The reason for this is, that the `MessageHandlerWidget`, which implements the handling of this message, is lazy initialized. Because of this the message handling dialog only appears if the "Messages" tab was previously selected in the tool selector. I could think of two ways around this:
   * Force the initialization of the `MessageHandlerWidget`. This would certainly solve the problem, but feels dirty: for later readers of the code it would not be appearent why this is the only widget that is not lazy initialized.
   * Move the error handling code to somewhere else. This seemed to be the cleaner way to go, so I did this. I extracted thi implementation fo the message handling dialog into its own class, and moved the invocation code into the `MainWindow`. This also has the side effect that I can set the main window as the parent for the message handling dialog, which this way receives the GammaRay icon, improving user experience :)

(These changes were also included here: https://github.com/KDAB/GammaRay/pull/112)


milianw:
In general, +1 - but this should be one commit, not four with one fixup and two merge commits. Could you clean that up please?

agabor:
Sure, no problem.

krf:
@agabor: No need to close this, you could've just updated your branch.

agabor:
@krf Thanks for the information, I didn't know that. I will do like that next time.

PR #145, opened 2015-08-20
Add handler for SIGHUP, SIGINT and SIGTERM.

akreuzkamp:
When receiving a signal to quit in the launcher, we want to quit the
debuggee as well. When hitting CTRL+C this happens, because it will
send SIGINT to the whole process group. But when only gammaray (a.k.a.
the launcher) receives the signal (e.g. killall -2 gammaray) the
debuggee won't quit.
The new handler simply calls QCoreApplication::quit, which causes the
client process to be destroyed.

### DIFF
--- a/launcher/main.cpp
+++ b/launcher/main.cpp
@@ -51,6 +51,40 @@
 #include <QStringList>
 #include <QVariant>
 
+#include <csignal>
+
+namespace {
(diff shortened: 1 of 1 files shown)


milianw:
lgtm

vkrause:
Looks good to me as well, feel free to integrate into master.

PR #146, opened 2015-09-01
Autosave on propertyextendededitor.

akreuzkamp:
So far for extended property editors, that open a dialog, the user had
to press Apply *and* press Enter afterwards to save the value.
Now it will commit the data as soon as the user pressed apply.

### DIFF
--- a/ui/propertyeditor/propertycoloreditor.cpp
+++ b/ui/propertyeditor/propertycoloreditor.cpp
@@ -41,7 +41,7 @@ void PropertyColorEditor::edit()
 {
   const QColor color = QColorDialog::getColor(value().value<QColor>(), this);
   if (color.isValid()) {
-    setValue(QVariant::fromValue(color));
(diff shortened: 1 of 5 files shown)


vkrause:
Thanks, general approach looks good. The manually enter event sending looks dirty though, that relies on specific keyboard behavior. Is there a nicer way to achieve this, looking e.g. at commitData() or closeEditor() in QAbstractItemDelegate?

akreuzkamp:
Yes, it's not the nicest way...
The alternative though, is to find a way for the extendedpropertyeditor to tell the delegate to save. Either the editor has to know the delegate (which I didn't see an easy way to do it) or it must emit a signal, which the delegate could react to, but that could easily result in dangling connections, making this approach quite error-prone...

vkrause:
Ok, then let's integrate it as-is, clearly a step forward.

PR #150, opened 2015-10-12
Make it possible to force different install RPATH through cmake cache

e4z9:
That makes it much easier to build and package GammaRay separately from the Qt that it is built with, without the need to fiddle around with rpaths after the fact.


CMakeLists.txt
```diff
@@ -37,8 +37,12 @@ endif()
...
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-set(CMAKE_MACOSX_RPATH TRUE)
+if(NOT DEFINED CMAKE_INSTALL_RPATH_USE_LINK_PATH)
+  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+endif()
+if(NOT DEFINED CMAKE_MACOSX_RPATH)
+  set(CMAKE_MACOSX_RPATH TRUE)
+endif()
```
vkrause: These changes are missing in your patch for master, not needed there/not needed in general/accidentally lost?

I'd usually just integrate the 2.3 patch and merge to master, unless we actually need different behavior here.

e4z9: Hm, good point. no idea what happened there. I'm fine with integrating to 2.3 and merging (and canceled the master one now)

vkrause:
Integrated into 2.3 and master now, thanks!

PR #153, opened 2015-10-20
Add code navigation feature for QtQuick Items.

akreuzkamp:
In quick inspector you can now right click an Item from the object
tree and click on "Show Code". This will open the source file, where
the item was declared, in the IDE selected via
Settings->Code Navigation. It also supports overriding the default
open action in order to integrate with an IDE directly (e.g. Qt
Creator).

### DIFF
--- a/ui/mainwindow.cpp
+++ b/ui/mainwindow.cpp
@@ -33,6 +33,7 @@
 #include "aboutdialog.h"
 #include "clienttoolmodel.h"
 #include "aboutdata.h"
+#include "uiintegration.h"
 
 #include "common/objectbroker.h"
 #include "common/modelroles.h"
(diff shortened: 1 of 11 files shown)


ui/mainwindow.cpp
```diff
@@ -48,12 +49,39 @@
...
 using namespace GammaRay;
+struct IdeSettings {
+    const char* const app;
+    const char* const args;
+    const char* const name;
+    const char* const icon;
+};
+
+static const IdeSettings ideSettings[] {
+#if defined(Q_OS_WIN) || defined(Q_OS_WIN)
```
vkrause: Q_OS_WIN twice?

ui/mainwindow.cpp
```diff
@@ -127,6 +155,79 @@ MainWindow::MainWindow(QWidget *parent): QMainWindow(parent), ui(new Ui::MainWin
...
+  action = new QAction(menu);
+  action->setText(QObject::tr("Automatic (No Line numbers)"));
+  action->setCheckable(true);
+  action->setChecked(currentIdx == -2);
+  action->setData(-2);
+  group->addAction(action);
+  menu->addAction(action);
+#endif
+
+  QObject::connect(group, &QActionGroup::triggered, [parent](QAction *action) {
```
vkrause: This wont work with Qt 4.8 I'm afraid.

vkrause:
Looks good in general, feel free to integrate this once the complains of the CI have been addressed.

PR #154, opened 2015-11-10
Rewrite of the Quick Scene Preview.

akreuzkamp:
Until now the quick inspector's preview widget was written in QML and
using QtQuick. It used Rectangles and Repeaters to draw a ruler,
got the geometry data as a QVariantMap and had some more dirty hacks
to function. Moreover it simply didn't work reliably (e.g. it used to
crash, when zooming). Last but not least, it was not stylable using Qt
Themes because of not using QWidget.

Now, the widget was rewritten as a custom QWidget. That makes the
preview
- much faster
- rock solid (at least better than before ;) )
- themable (uses Manhattan style when embedded into QtCreator)
- better maintainable

The code which draws the overlays about an items geometry didn't
change, but was just adopted one by one.


vkrause:
Looks good to me, feel free to integrate if the CI has no objections.

PR #155, opened 2015-11-10
dot.exe does not print its version -> removed OUTPUT_QUIET flag for 'process command'

ChristopherHahn:
cmake configuration failed on Windows because dot.exe did not print its version


winterz:
coincidentally, I ran into the same problem on another project this morning.  I had a slightly different solution, but I like your solution better. I'll test your solution in the other project first.

winterz:
thanks

PR #161, opened 2015-11-27
Fix wrong destruction of MethodArgument.

akreuzkamp:
The destructor of MethodArgument used type() to determine the type of
the QVariant stored in it. But type is QMetaType::UserType for all
user types. Thus it would try to call the destructor of QWidget*,
which is the first user type. That causes the value not to be
destructed and thus caused a memory leak, causing thousands of images
not to be destructed.
Now it uses userType() to correctly get the actual type
(userType() == type() for non-user types).


milianw:
ouch! lgtm - feel free to merge

vkrause:
indeed, surprising this never hit us before

PR #168, opened 2016-01-13
Install mkspecs stuff inside the bundle resources on OSX.

pasnox:
The mkspecs/modules folder was outside the bundle which is not good.


vkrause:
looks good to me, feel free to integrate this into master

PR #169, opened 2016-01-13
Make Qt layout work on OSX.

pasnox:
Only the default probe(s) is/are handled.
Handling multiple probes may appear later.


vkrause:
looks good to me, feel free to integrate this into master

PR #170, opened 2016-01-13
Handle multiple probes for Qt layout builds.

pasnox:
Task-Id: KDEND-34


vkrause:
looks good to me, thanks!

PR #171, opened 2016-01-14
Fix bad iterator dereference in RemoteViewWidget::sourceTickLabelDist…

dereksbg:
…ance()


winterz:
volker, can you take a look?

vkrause:
Yep, that's indeed correct, I'll integrate it, thanks!

PR #172, opened 2016-01-21
OSX now has its probes splitted like windows.

pasnox:
OSX do require separates debug/release probes to avoid mixed Qt version
errors at run time.


vkrause:
Besides the issues we discussed yesterday already, the CI also noticed unit test failures.

pasnox:
Ok, forget this pull until I do give comment feedback here.

PR #173, opened 2016-01-21
Add property editors for various matrix classes

tokoe-kdab:
Add property editors for
  - QMatrix
  - QMatrix4x4
  - QTransform
  - QVector2D
  - QVector3D
  - QVector4D


vkrause:
Looks good to me in general, feel free to integrate into master. Some of these types (QVectorXD) are not available in all supported Qt versions though (see propertyeditordelegate.cpp), if you push to a branch in the main repo rather than in your clone the CI will verify that to be sure.

Another thing I noticed is that there is a bit of code duplication with matrix_trait in propertyeditordelegate.cpp, but that is compile-time type selection rather than runtime, so might not be worth it.

PR #177, opened 2016-01-26
MainWindow: Handle tool buttons in tool's actions

krf:


krf:
Main use case: http://imgur.com/Dxhu491

krf:
(10:45:30 AM) volker: #177 looks sane to me

PR #179, opened 2016-01-26
Delete cached widgets

bog-dan-ro:


bog-dan-ro:
Fixed clear function

krf:
Pushed to 2.4, with minor changes.

Thanks!

PR #182, opened 2016-01-29
Global context menu support

krf:
Changes:
* ProbeController: Allow to select objects (QObjects-only atm)
* ProbeController: Allow to retrieve supports tools for a object
  (QObject-only)
* Introduce a (remotable) ObjectIdRole in ObjectModelBase
* Introduce ObjectId, a class storing a quint64 representing a raw
  pointer on the debuggee
* Introduce a ContextMenuExtension class, which can be used on client
  QAIVs (actually, wherever we may show context menus)
  * Input: ObjectId
  * Output: Actions added to a QMenu*
  * Presents available tools for the moment, nothing else

Only implemented for tech-preview for Object Inspector and Properties
Extension views

TODO:
* Support selecting non-QObjects again (needs more API in
  ProbeController)
* Figure out why we can't make pass ObjectId from server->client
  (currently transported as raw quint64)
* Limit the amount of tools shown (currently almost every tool is shown,
  as they support "QObject" (which every QObject inherits from,
  obviously)
* Split up in multiple commits

Task-Id: KDSME-10

### DIFF
--- a/common/probecontrollerinterface.h
+++ b/common/probecontrollerinterface.h
@@ -30,15 +30,80 @@
 #define GAMMARAY_PROBECONTROLLERINTERFACE_H
 
 #include <QObject>
+#include <QDataStream>
+#include <QVector>
 
 namespace GammaRay {
 
-/** Probe and host process remote control functions. */
-class ProbeControllerInterface
+class ObjectId
 {
 public:
-  ProbeControllerInterface();
+  enum Type
+  {
+    Invalid,
+    QObjectType,
+    VoidStarType
+  };
+
+  explicit ObjectId(void *obj, const QByteArray &typeName)
+    : m_type(VoidStarType)
+    , m_id(reinterpret_cast<quint64>(obj))
+    , m_typeName(typeName)
+  {}
+  explicit ObjectId(QObject *obj)
+    : m_type(QObjectType)
+    , m_id(reinterpret_cast<quint64>(obj))
+  {}
+  explicit ObjectId()
+    : m_type(Invalid)
+    , m_id(0)
+  {}
+  inline bool isNull() const { return m_id == 0; }
+  inline quint64 id() const { return m_id; }
+  inline Type type() const { return m_type; }
+  inline QByteArray typeName() const { return m_typeName; }
+
+  inline QObject *asQObject()
+  {
+    Q_ASSERT(m_type == QObjectType);
+    return reinterpret_cast<QObject *>(m_id);
+  }
+  inline void *asVoidStar()
+  {
+    Q_ASSERT(m_type == VoidStarType);
+    return reinterpret_cast<void *>(m_id);
+  }
+
+  inline operator quint64() const { return m_id; }
+
+private:
+  friend QDataStream &operator<<(QDataStream &out, ObjectId id);
+  friend QDataStream &operator>>(QDataStream &out, ObjectId &id);
+
+  Type m_type;
+  quint64 m_id;
+  QByteArray m_typeName;
+};
+
+struct ToolInfo
+{
+  QString id;
+  QString name;
+};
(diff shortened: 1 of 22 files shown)


krf:
Patch updated, good to go now.

common/probecontrollerinterface.h
```diff
@@ -30,28 +30,132 @@
...
+  in >> toolInfo.id;
+  in >> toolInfo.name;
+  return in;
+}
+
 }
 
 Q_DECLARE_INTERFACE(GammaRay::ProbeControllerInterface, "com.kdab.GammaRay.ProbeControllerInterface")
+Q_DECLARE_METATYPE(GammaRay::ObjectId)
+Q_DECLARE_TYPEINFO(GammaRay::ObjectId, Q_PRIMITIVE_TYPE);
```
vkrause: Are you sure it's still a primitive type with the QByteArray member?

ui/tools/objectinspector/objectinspectorwidget.cpp
```diff
@@ -76,3 +87,18 @@ void ObjectInspectorWidget::objectSelectionChanged(const QItemSelection& selecti
...
 }
+
+void ObjectInspectorWidget::objectContextMenuRequested(const QPoint& pos)
+{
+  const auto index = ui->objectTreeView->indexAt(pos);
+  if (!index.isValid())
+    return;
+
+  const auto objectId = index.data(ObjectModel::ObjectIdRole).value<ObjectId>();
+  qWarning() << "FOO" << objectId;
```
vkrause: left-over debug output?

core/objectmodelbase.h
```diff
@@ -59,6 +60,8 @@ class ObjectModelBase : public Base
      */
     explicit ObjectModelBase<Base>(QObject *parent) : Base(parent)
     {
+      qRegisterMetaType<ObjectId>();
+      qRegisterMetaTypeStreamOperators<ObjectId>();
```
vkrause: Are those needed here? Looks like the ProbeControllerInterface does this already?

ui/tools/objectinspector/objectinspectorwidget.cpp
```diff
@@ -44,11 +48,15 @@ ObjectInspectorWidget::ObjectInspectorWidget(QWidget *parent)
   : QWidget(parent),
     ui(new Ui::ObjectInspectorWidget)
 {
+  qRegisterMetaType<ObjectId>();
+  qRegisterMetaTypeStreamOperators<ObjectId>();
```
vkrause: same as above?

ui/tools/objectinspector/propertiestab.cpp
```diff
@@ -58,6 +59,9 @@ PropertiesTab::PropertiesTab(PropertyWidget *parent)
   m_ui->newPropertyButton->setIcon(QIcon::fromTheme(QStringLiteral("list-add")));
 
   setObjectBaseName(parent->objectBaseName());
+
+  qRegisterMetaType<ObjectId>();
+  qRegisterMetaTypeStreamOperators<ObjectId>();
```
vkrause: same as above

vkrause:
Looks good in general, feel free to integrate, definitely a big step into the right direction :)

krf:
I fear the `qRegisterMetaType{,StreamOperator}` calls are still necessary, otherwise GammaRay crashes with:

```
[gammaray-client(16858)/(default) unknown(0): ASSERT failure in QVector<T>::at: "index out of range", file /usr/include/x86_64-linux-gnu/qt5/QtCore/qvector.h, line 393
```

=> Not able to deserialize ObjectId on the client side, due to not being aware of it yet.

krf:
I'll push now, we can remove those when we find a better solution (I can't think of any) in the future.

PR #186, opened 2016-03-03
Fix unit tests with Qt 4.8.

pasnox:
This is done porting the debug postfix to Qt4 too.
So now both Qt4 & Qt5 ports are consistent with loading plugins.

Task-Id: KDEND-48


pasnox:
Merged.

PR #195, opened 2016-03-22
Fix selection synchronization on server-side sorting/filtering.

pasnox:
The server is now maintaining a 'at least one row selected' per model
via the NSM::requestSelection().
The UI part should no longer set explicit default selection on its own.

Task-Id: KDEND-43

### DIFF
--- a/client/selectionmodelclient.cpp
+++ b/client/selectionmodelclient.cpp
@@ -29,11 +29,25 @@
 #include "selectionmodelclient.h"
 #include "client.h"
 
+#include <QTimer>
+
 using namespace GammaRay;
 
 SelectionModelClient::SelectionModelClient(const QString& objectName, QAbstractItemModel* model, QObject* parent) :
-  NetworkSelectionModel(objectName, model, parent)
+  NetworkSelectionModel(objectName, model, parent), m_timer(new QTimer(this))
 {
+  m_timer->setSingleShot(true);
+  m_timer->setInterval(125);
+  Q_ASSERT(model);
+  // We do use a timer to group requests to avoid network overhead
+  connect(model, SIGNAL(modelAboutToBeReset()), this, SLOT(clearPendingSelection()));
+  connect(model, SIGNAL(rowsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(rowsMoved(QModelIndex,int,int,QModelIndex,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsMoved(QModelIndex,int,int,QModelIndex,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(layoutChanged()), m_timer, SLOT(start()));
+  connect(m_timer, SIGNAL(timeout()), this, SLOT(timeout()));
+
   m_myAddress = Client::instance()->objectAddress(objectName);
   connect(Client::instance(), SIGNAL(objectRegistered(QString,Protocol::ObjectAddress)), SLOT(serverRegistered(QString,Protocol::ObjectAddress)));
   connect(Client::instance(), SIGNAL(objectUnregistered(QString,Protocol::ObjectAddress)),  SLOT(serverUnregistered(QString,Protocol::ObjectAddress)));
@@ -42,6 +56,7 @@ SelectionModelClient::SelectionModelClient(const QString& objectName, QAbstractI
 
 SelectionModelClient::~SelectionModelClient()
 {
+  m_timer->stop();
 }
 
 void SelectionModelClient::connectToServer()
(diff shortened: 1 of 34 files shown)


pasnox:
There is still possible issues, but require some discussion:
- When the network selection model do receive a message and it is already busy handling one, the message get skipped.
- When the selection model is created ui wise, messages can be received in between the selection model creation and its connection, the message is then lost.
Maybe we may need to message pool here to make sure some message get call at a later time ? (like the requestSelection one).

Comments, ideas, welcomes.

client/selectionmodelclient.cpp
```diff
@@ -29,11 +29,22 @@
...
-  NetworkSelectionModel(objectName, model, parent)
+  NetworkSelectionModel(objectName, model, parent), m_timer(new QTimer(this))
 {
+  m_timer->setSingleShot(true);
+  m_timer->setInterval(125);
+  Q_ASSERT(model);
+  // We do use a timer to group requests to avoid network overhead
+  connect(model, SIGNAL(modelAboutToBeReset()), this, SLOT(clearPendingSelection()));
+  connect(model, SIGNAL(rowsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
```
vkrause: Shouldn't layoutChanged be in this list instead, row/column changes should automatically update persistent model indexes, and thus selections, no?

common/networkselectionmodel.cpp
```diff
@@ -43,87 +49,166 @@ static void writeSelection(Message *msg, const QItemSelection &selection)
...
+    return;
+  }
+
+  clearPendingSelection();
+
+  const QItemSelectionModel::SelectionFlags flags = QItemSelectionModel::ClearAndSelect |
+      QItemSelectionModel::Rows | QItemSelectionModel::Current;
+
+  if (!hasSelection()) {
+    // Only match tools selection: com.kdab.GammaRay.XXX.selection
```
vkrause: The following code looks like it's in the wrong place here. The basic communication infrastructure should not depend on specific tools. Can't we find a way to move this logic into the corresponding plug-in? If we need to pass information into the selection model from the plug-ins, worst case we can do this with dynamic properties, which wouldn't be any less fragile than this approach I think.

client/selectionmodelclient.cpp
pasnox: Indeed layoutChanged may also be there.

common/networkselectionmodel.cpp
pasnox: Right, I will find a way to put this back to the corresponding plugin for specific case.
Though selecting index(0,0) can stay there i guess (generic) ?

common/networkselectionmodel.cpp
```diff
@@ -43,87 +46,174 @@ static void writeSelection(Message *msg, const QItemSelection &selection)
...
+
+  clearPendingSelection();
+
+  const QItemSelectionModel::SelectionFlags selectionFlags = QItemSelectionModel::ClearAndSelect |
+      QItemSelectionModel::Rows | QItemSelectionModel::Current;
+  const Qt::MatchFlags matchFlags = Qt::MatchExactly | Qt::MatchRecursive;
+
+  if (!hasSelection()) {
+    // Only match tools selection: com.kdab.GammaRay.XXX.selection
+    if (QRegExp(QStringLiteral("com\\.kdab\\.GammaRay\\.[\\w]+\\.selection")).exactMatch(m_objectName)) {
```
vkrause: This looks very fragile, as there is no clear policy on how objects are named. It's for example not matching the entity and frame graph node trees of the Qt3D inspector, which are candidates for default selection.

Isn't the following check for defaultSelectedItem already excluding anything that doesn't want default selection? If not, could we do this in a similar way, e.g. have a dynamic property to indicate no default selection, should this actually be needed somewhere?

common/protocol.h
```diff
@@ -84,6 +85,7 @@ enum BuildInMessageType {
   ModelColumnsRemoved,
   ModelReset,
   ModelLayoutChanged,
+  SelectionModelClearSelect,
```
vkrause: Better bump the protocol version as well, when changing things in here.

core/remote/selectionmodelserver.cpp
```diff
@@ -29,16 +29,37 @@
...
+  m_timer->setSingleShot(true);
+  m_timer->setInterval(125);
+  Q_ASSERT(model);
+  // We do use a timer to group requests to avoid network overhead
+  connect(model, SIGNAL(rowsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsInserted(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(rowsMoved(QModelIndex,int,int,QModelIndex,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(rowsRemoved(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsRemoved(QModelIndex,int,int)), m_timer, SLOT(start()));
+  connect(model, SIGNAL(columnsMoved(QModelIndex,int,int,QModelIndex,int)), m_timer, SLOT(start()));
```
vkrause: Do we really need all of them? Persistent model indexes (and thus selections) should be automatically updated on insert/move/remove, no?

common/protocol.h
pasnox: Done

core/remote/selectionmodelserver.cpp
pasnox: I guess i could remove the XXXRemoved one.
Others might be needed, in fact the slots slotCurrentChanged / slotCurrentColumnChanged / slotCurrentRowChanged does not propagate original passed flags, but hard fixed one.
This can lead to 'lost' flags, like the 'Clear' one etc, unfortunately not all members in the QISM are virtual in Qt4, so we can't override them all.
The best we an do here is to request to send again the correct selection.

common/networkselectionmodel.cpp
pasnox: A model that does not define the defaultSelectedItem would have its 0,0 index selected.
I will remove the check and see if no problems appear as is.

core/remote/selectionmodelserver.cpp
vkrause: Ah, so Qt4 compatibility is why we can't use the virtuals in QISM? I was wondering already, that should avoid the flag issue. Ok, then we probably don't have a choice here.

pasnox:
Merged

PR #199, opened 2016-03-31
Open qrc navigation in our Resource Inspector.

pasnox:
Task-Id: KDEND-40

### DIFF
--- a/common/resourcebrowserinterface.h
+++ b/common/resourcebrowserinterface.h
@@ -44,11 +44,12 @@ class ResourceBrowserInterface : public QObject
 
   public slots:
     virtual void downloadResource(const QString &sourceFilePath, const QString &targetFilePath) = 0;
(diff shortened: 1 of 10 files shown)


akreuzkamp:
qrc code navigation currently only works with QtCreator (or theoretically with any ide, that supports it). Your idea is good, but it breaks the qrc code navigation in QtCreator, where we definitely want to navigate to the editor, doesn't it? I guess we'd need some way for the qtc-gammaray-plugin to change that behavior.

pasnox:
If I recall, now QtC does not longer use the client MainWindow ?
So we can write our own handler.
If not then maybe we can rework a bit the api to add a virtual member somewhere.

akreuzkamp:
you're right, sorry. QtC never used MainWindow::navigateToCode(), so this patch doesn't in any way interfere with code navigation in QtC.

vkrause:
+2

vkrause:
Integrated into master.

PR #201, opened 2016-04-07
Implement code navigation for QObjects

akreuzkamp:
Finding the code location is done by generating a backtrace from Probe::objectAdded, which is called from the QObject constructor. From there we can step up to the frame above the object's constructor and get the source location of that frame.

The stacktrace generation is based on backward-cpp.

It has a few issues I know of:
- ~~outputs a whole lot of warnings coming from backward. I'm not sure, how to fix/silence them.~~
- There is no handling for the case we want to show multiple creation locations (i.e. qml- and C++ location). But I think it's ok for now. We can add that, when it's needed.
- It seems, that the code locations are fetched for every object, not only on right click (as I implemented it before). Maybe someone (Milian? :D) can check for performance issues before we ship it. :)
- ~~backward.hpp has a variable called "signals" (guess, what happens :D). I've created a pull request to rename it, but right now we need to include it using~~

``` C++
#pragma push_macro("signals")
#undef signals
#include <3rdparty/backward-cpp/backward.hpp>
#pragma pop_macro("signals")
```

~~which is non-standard afaik, but it seems to work with all CI-covered compilers.~~

### DIFF
--- a/core/probe.cpp
+++ b/core/probe.cpp
@@ -53,6 +53,8 @@
 #include <common/streamoperators.h>
 #include <common/paths.h>
 
+#include <3rdparty/backward-cpp/backward.hpp>
+
 #if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
 #include <QApplication>
 #else
@@ -194,6 +196,7 @@ struct Listener
 
     bool trackDestroyed;
     QVector<QObject *> addedBeforeProbeInstance;
+    QHash<QObject *, backward::StackTrace> constructionBacktracesForObjects;
 };
 
 Q_GLOBAL_STATIC(Listener, s_listener)
@@ -209,6 +212,7 @@ Probe::Probe(QObject *parent)
     , m_window(nullptr)
     , m_queueTimer(new QTimer(this))
     , m_server(nullptr)
+    , m_traceResolver(new backward::TraceResolver())
 {
     Q_ASSERT(thread() == qApp->thread());
     IF_DEBUG(cout << "attaching GammaRay probe" << endl;)
@@ -537,6 +541,9 @@ void Probe::objectAdded(QObject *obj, bool fromCtor)
 
 #endif
 
+    if (fromCtor) {
+        s_listener()->constructionBacktracesForObjects[obj].load_here(32);
+    }
     if (!isInitialized()) {
         IF_DEBUG(cout
                  << "objectAdded Before: "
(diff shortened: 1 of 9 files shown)


akreuzkamp:
Update: The pull request concerning usage of signals was accepted and merged. I updated this pull request to reflect this change. :)

vkrause:
Doesn't build in BFD mode, but DW mode works fine, and works so far very fast and precise, impressive :) Still need to test without debug builds and with higher QObject loads though. Anyway, already identified a QObject we are leaking from GammaRay :)

core/probe.cpp
```diff
@@ -1015,3 +1021,41 @@ void Probe::executeSignalCallback(const Func &func)
...
+  }
+
+  backward::StackTrace &st = s_listener()->constructionBacktracesForObjects[object];
+  int distanceToQObject = 0;
+  m_traceResolver.load_stacktrace(st);
+
+  const QMetaObject *metaObject = object->metaObject();
+  while (metaObject != &QObject::staticMetaObject) {
+    distanceToQObject++;
+    metaObject = metaObject->superClass();
```
vkrause: This needs a nullptr check somewhere, as there are (rare) QObject instances whose super class chain does not terminate in a QObject (value type adaptors for QML for example).

krf:
Note: Please pack the introduction of the 3rdparty lib into a separate commit

akreuzkamp:
ping :)

akreuzkamp:
ping :)

krf:
Rebased & added another patch to disable backtrace logging via an env var

krf:
Measured the performance impact of the backtrace logging, by starting QtCreator in GammaRay:

Stats with feature *disabled* (`GAMMARAY_DISABLE_BACKTRACE_LOGGING=1` set):
```
 Performance counter stats for 'gammaray --inject-only /home/kfunk/devel/build/qt-creator-4.2/bin/qtcreator':

      11505,077845      task-clock (msec)         #    0,937 CPUs utilized          
             3.489      context-switches          #    0,303 K/sec                  
               251      cpu-migrations            #    0,022 K/sec                  
            67.747      page-faults               #    0,006 M/sec                  
    36.491.745.829      cycles                    #    3,172 GHz                    
    55.802.425.291      instructions              #    1,53  insn per cycle         
     8.914.173.959      branches                  #  774,803 M/sec                  
        32.899.284      branch-misses             #    0,37% of all branches        

      12,280188511 seconds time elapsed
```

Stats with feature *enabled*:

```
 Performance counter stats for 'gammaray --inject-only /home/kfunk/devel/build/qt-creator-4.2/bin/qtcreator':

      11910,462233      task-clock (msec)         #    0,965 CPUs utilized          
             2.968      context-switches          #    0,249 K/sec                  
               270      cpu-migrations            #    0,023 K/sec                  
            71.894      page-faults               #    0,006 M/sec                  
    37.876.512.429      cycles                    #    3,180 GHz                    
    57.614.118.337      instructions              #    1,52  insn per cycle         
     9.344.738.286      branches                  #  784,582 M/sec                  
        38.801.024      branch-misses             #    0,42% of all branches        

      12,339536205 seconds time elapsed
```

Instruction count goes up from 55.802 M to 57.614 M => 3 % slow down.

Did the same measurements on a simpler example (traffic), that gave me a 5-6% slow down.

So I think we can expect something like a 3-6% slow down with this feature enabled. Using GammaRay on e.g. QtCreator with the feature enable still feels "okay" though.

krf:
Patch LGTM, I think we can merge it. It's definitely a pretty cool feature GammaRay should have.

You can disable the feature by setting the following env var now:
```
GAMMARAY_DISABLE_BACKTRACE_LOGGING=1
```

krf:
@akreuzkamp Another feature/bug request: We should make sure we don't attempt to open non-existing files (sometimes GammaRay suggests I can open e.g. `.moc/main.moc` (which is not retrievable))

akreuzkamp:
mhh, not sure about your request...It's simple to add a check if the file exists, but especially for relative paths it would prevent IDEs to find those files. Although currently no IDE does it (except qtcreator in plugin-mode, but that's separated anyway), they could search their current project paths for the file and find it. Unless we find a way to resolve paths ourselves and be sure, that .moc/main.moc really doesn't exist anymore.

PR #210, opened 2016-05-02
Fix coding style

krf:


winterz:
compiles and the tests are ok.
I wonder if you can also enforce line length of 100 chars ?

krf:
@vkrause: Needs rebase -- but generally: Are you fine with this change? Any additions to the coding style rules needed?

krf:
Reran my scripts, updated the commit message. Unit tests still work.

vkrause:
qt3dgeometryextensioninterface.cpp looks messed up, the rest is ok I think, so with that fixed this can go in.

krf:
I fixed *qt3dgeometryextensioninterface.cpp*, for what it's worth. Uncrustify messed that up indeed. Thanks for the hint.

PR #214, opened 2016-05-09
CMake: Better error message if both Qt4+5 missing

### LINKED ISSUE #198: Nicer CMake messages if neither Qt4 nor Qt5 are found
refer to  https://github.com/KDAB/GammaRay/issues/197
if neither Qt4 nor Qt5 can be located we should print a nice message about that; not say that Qt4 cannot be found.

krf:
Fixes #198


winterz:
looks good. tested on Windows where I don't have any qmake in my %PATH%

PR #218, opened 2016-05-24
Use the property editor delegate in the model inspector.

pasnox:
As this inspector used special delegate to avoid empty cells,
the delegate code was splitted to be merged into PropertyEditorDelegate.


pasnox:
Don't merge yet - just seen that now we can no longer edits models values - probably something related to data types.

vkrause:
+2

pasnox:
Merged

PR #219, opened 2016-05-24
Fix incorrect editor height on OSX.

pasnox:
On OSX when a button is used in a layout, its sibling widgets
get extra margin, reducing their heights.
In this case, the label height was too small to be readable.

To break this weirdness layout behavior on OSX, we need to put the
button in a dedicated widget.


vkrause:
+2

pasnox:
Merged

PR #220, opened 2016-05-24
Introduce multiline editor for String properties

pasnox:
This was a lot annoying editing long text properties.

There is now a multiline text editor for those properties.
This also fix the tree view rows height being too big because of those texts.


pasnox:
Merged

PR #222, opened 2016-05-26
Rework cmake files

pasnox:
This set the include paths for each modules in
the installed cmake files and allow external plugins to
easily link gammaray.

Probe only building now no longer install development files (headers and cmake files).

Task-Id: KDEND-61


krf:
:+1:

PR #223, opened 2016-05-31
List picking in RemoteView support

pasnox:
A new mode appearing allowing picking an item from a tree list.
This mode is optional and triggered with a keyboard Shift+Control modifier while clicking on the preview.
The old one click picking has been reworked to use the new underlying item discovery and is still the default.

### DIFF
--- a/common/remoteviewinterface.h
+++ b/common/remoteviewinterface.h
@@ -31,6 +31,8 @@
 
 #include "gammaray_common_export.h"
 
+#include "objectid.h"
+
 #include <QObject>
 #include <QPoint>
 
@@ -42,12 +44,18 @@ class GAMMARAY_COMMON_EXPORT RemoteViewInterface : public QObject
 {
     Q_OBJECT
 public:
+    enum RequestMode {
+        RequestBest,
+        RequestAll
+    };
+
     explicit RemoteViewInterface(const QString &name, QObject *parent = Q_NULLPTR);
 
     QString name() const;
 
 public slots:
-    virtual void pickElementAt(const QPoint &pos) = 0;
+    virtual void requestElementsAt(const QPoint &pos, GammaRay::RemoteViewInterface::RequestMode mode) = 0;
+    virtual void pickElementId(const GammaRay::ObjectId &id) = 0;
 
     virtual void sendKeyEvent(int type, int key, int modifiers,
                               const QString &text = QString(), bool autorep = false,
@@ -66,12 +74,16 @@ public slots:
 
 signals:
     void reset();
+    void elementsAtReceived(const GammaRay::ObjectIds &ids, int bestCandidate);
     void frameUpdated(const GammaRay::RemoteViewFrame &frame);
 
 private:
     QString m_name;
 };
+
 }
+
+Q_DECLARE_METATYPE(GammaRay::RemoteViewInterface::RequestMode)
 
 QT_BEGIN_NAMESPACE
 Q_DECLARE_INTERFACE(GammaRay::RemoteViewInterface, "com.kdab.GammaRay.RemoteViewInterface/1.0")
--- a/plugins/quickinspector/quickinspector.cpp
+++ b/plugins/quickinspector/quickinspector.cpp
@@ -223,7 +223,7 @@ static QString qsgTextureWrapModeToString(QSGTexture::WrapMode wrapMode)
 
 static bool isGoodCandidateItem(QQuickItem *item)
 {
-    if (!item->isVisible() || qFuzzyCompare(item->opacity() + 1.0, qreal(1.0)) ||
+    if (!item->isVisible() || item->z() < 0 || qFuzzyCompare(item->opacity() + 1.0, qreal(1.0)) ||
             !item->flags().testFlag(QQuickItem::ItemHasContents) ||
             item->metaObject() == &QQuickItem::staticMetaObject) {
         return false;
@@ -616,7 +616,7 @@ ObjectIds QuickInspector::recursiveItemsAt(QQuickItem *parent, const QPointF &po
                 int bc;
                 objects << recursiveItemsAt(c, p, mode, bc);
 
-                if (bestCandidate == -1 && bc != -1) {
+                if (bestCandidate == -1 && bc != -1 && c->z() >= 0) {
                     bestCandidate = count + bc;
                 }
             }
--- a/plugins/quickinspector/quickinspector.cpp
+++ b/plugins/quickinspector/quickinspector.cpp
@@ -35,6 +35,7 @@
 
 #include <common/modelevent.h>
 #include <common/objectbroker.h>
+#include <common/probecontrollerinterface.h>
 #include <common/remoteviewframe.h>
 
 #include <core/metaobject.h>
@@ -220,6 +221,17 @@ static QString qsgTextureWrapModeToString(QSGTexture::WrapMode wrapMode)
     return QStringLiteral("Unknown: %1").arg(wrapMode);
 }
 
+static bool isGoodCandidateItem(QQuickItem *item)
+{
+    if (!item->isVisible() || qFuzzyCompare(item->opacity() + 1.0, qreal(1.0)) ||
+            !item->flags().testFlag(QQuickItem::ItemHasContents) ||
+            item->metaObject() == &QQuickItem::staticMetaObject) {
+        return false;
+    }
+
+    return true;
+}
+
 QuickInspector::QuickInspector(ProbeInterface *probe, QObject *parent)
     : QuickInspectorInterface(parent)
     , m_probe(probe)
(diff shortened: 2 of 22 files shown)


plugins/quickinspector/quickinspector.cpp
```diff
@@ -228,6 +229,17 @@ static QString qsgTextureWrapModeToString(QSGTexture::WrapMode wrapMode)
   return QStringLiteral("Unknown: %1").arg(wrapMode);
 }
 
+static bool qGoodCandidateItem(QQuickItem *item)
```
hasselmm: the "q" prefix indicates to the incautious reader, that Qt provides this function. is this intentional?

pasnox: Just tried to stay in the line of the previous members which also prefix with q.
But it's absolutely not linked to Qt itself.

akreuzkamp:
Imho, this makes item picking too complicated and isn't worth it... I'd stick to Mathias' suggestion from the bug report, to just add some more intelligence to item picking (ie. filtering out invisible items)

But don't consider it as a veto, if the rest likes this approach, feel free. :)

vkrause:
It might very well be that this introduces a usability issue for the "normal" case of just wanting to pick the top visual item, that doesn't mean we should kill the feature though, it's very useful for more advanced scenarios, like for looking into event handling issues.

Maybe we need a "superficial pick" and a "full-depth pick" mode?

hasselmm:
Showing a list really should just be an option. By default GammaRay should just pick whatever item that got accepted first via that qBestCandidate() function.

plugins/quickinspector/quickinspector.cpp
hasselmm: Very strange, but I see.

plugins/quickinspector/quickinspector.cpp
```diff
@@ -561,47 +574,71 @@ void QuickInspector::sgNodeDeleted(QSGNode *node)
...
       }
+    }
   }
 
-  if (child) {
-    return recursiveChiltAt(child, parent->mapToItem(child, pos));
+  objects << ObjectId(parent);
+
+  if (bestCandidate == -1 && qGoodCandidateItem(parent)) {
+    bestCandidate = objects.count();
```
hasselmm: In the loop you want objects.count() as you add to objects after updating bestCandidate, but in this case the objects list is fully setup already. So you are off by one.

hasselmm:
Doesn't seem to work. Checkout the m*-diagnosis project from our Gerrit, build with Qt 5.6 and try GammaRay on it. GammaRay stil proposes the hidden password overlay, instead of the picked images.

plugins/quickinspector/quickinspector.cpp
pasnox: Good catch, fixed.

pasnox:
Ok, I will make the tree picking an option (Will need a modifier key to be activated).
@hasselmm I do use neptune as my reference qml ui, can u reproduce issues here too ?

pasnox:
@hasselmm With the latests rebased changes, the hidden overlay is no longer selected here.
Can u confirm it does works correctly now ?

hasselmm:
Nope, still doesn't work. Now text labels and input fields of the hidden password overlay are picked:

QQuickRootItem/Item/loginOverlay/*

Instead children of the BluetoothStatus item should be picked:

QQuickRootItem/Item/mainView/BluetoothStatus/*

pasnox:
Just pushed latest changes.
Fixed building with Qt < 5.6, crash and quickinspector test.

@hasselmm : Please add steps to reproduce your problem.

pasnox:
Just added a new commit for z order, seems to make it.
Please feedback.

pasnox:
@hasselmm: ping

plugins/quickinspector/quickinspector.cpp
vkrause: That's because the other functions there are all of the form qTypeToString(), for a set of Qt types. So the "q" prefix for qGoodCandidateItem() indeed doesn't make sense.

common/probecontrollerinterface.h
```diff
@@ -146,6 +164,27 @@ inline QDataStream &operator>>(QDataStream &in, ObjectId &id)
   return in;
 }
 
+inline QDataStream &operator<<(QDataStream &out, const ObjectIds &ids)
```
vkrause: There is GAMMARAY_ENUM_STREAM_OPERATORS to avoid this boilerplate code.

vkrause: This looks like the default stream operators for QVector, why do we need an explicit definition of them?

common/remoteviewinterface.h
```diff
@@ -67,15 +75,31 @@ public slots:
...
 signals:
     void reset();
+    void elementsAtReceived(const GammaRay::ObjectIds &ids, int bestCandidate);
     void frameUpdated(const GammaRay::RemoteViewFrame &frame);
 
 private:
     QString m_name;
 };
 
+inline QDataStream &operator<<(QDataStream &out, RemoteViewInterface::RequestMode mode)
```
vkrause: See GAMMARAY_ENUM_STREAM_OPERATORS to avoid this boilerplate code.

core/objectidfilterproxymodel.h
```diff
@@ -0,0 +1,98 @@
...
+#include "objectmodelbase.h"
+
+#include <QSortFilterProxyModel>
+
+namespace GammaRay {
+
+/**
+ * @brief A QSortFilterProxyModel for generic ObjectIds.
+ */
+class GAMMARAY_CORE_EXPORT ObjectIdFilterProxyModelBase : public QSortFilterProxyModel
```
vkrause: Why is this a separate class, rather than being part of ObjectIdFilterProxyModel?

core/objectidfilterproxymodel.cpp
```diff
@@ -0,0 +1,88 @@
...
+        m_sorting[id.id()] = i;
+    }
+    invalidateFilter();
+}
+
+bool ObjectIdsFilterProxyModel::lessThan(const QModelIndex &source_left, const QModelIndex &source_right) const
+{
+    const GammaRay::ObjectId leftId = source_left.data(ObjectModel::ObjectIdRole).value<GammaRay::ObjectId>();
+    const GammaRay::ObjectId rightId = source_right.data(ObjectModel::ObjectIdRole).value<GammaRay::ObjectId>();
+    return m_sorting.value(leftId.id()) < m_sorting.value(rightId.id());
```
vkrause: Why do we need unconditional sorting by ObjectId here, rather than sorting by whatever the user sees?

ui/CMakeLists.txt
```diff
@@ -122,7 +123,7 @@ endif(HAVE_PRIVATE_QT_HEADERS)
 
 target_link_libraries(gammaray_ui
   LINK_PUBLIC ${QT_QTCORE_LIBRARIES} ${QT_QTGUI_LIBRARIES} gammaray_common
-  LINK_PRIVATE gammaray_common_internal gammaray_kitemmodels
+  LINK_PRIVATE gammaray_common_internal gammaray_kitemmodels gammaray_core
```
vkrause: This looks wrong, what do we need from gammaray_core here? Probably should move to common then...

common/probecontrollerinterface.h
pasnox: Removed

common/remoteviewinterface.h
pasnox: Done

core/objectidfilterproxymodel.h
pasnox: Will change to a KRFPM.
Why it's separated is that I tried to match what is done in objecttypefilterproxymodel.h.
I can merge them if it's what you prefer.

core/objectidfilterproxymodel.cpp
pasnox: I want the model to be sorted exactly like the m_ids vector is.
Because this list is exactly what the recursive item discoverer generate and is accurate with what is seeing visually the user.
But i may just need to override ::sort() and do nothing :)

ui/CMakeLists.txt
pasnox: We need objectidfilterproxymodel.h which is in core because... was trying to match objecttypefilterproxymodel.h.
Will move it to common then.

pasnox: Hm the proxy do require objectmodelbase.h (for the model role: ObjectModel::ObjectIdRole)
What should I do here ?

core/objectidfilterproxymodel.h
vkrause: It's separated there due to the use of templates, no?

ui/CMakeLists.txt
vkrause: objecttypefilterproxymodel.h relies on object pointers, so that has to be in core (use in ui would not work), ObjectId based code can work on both sides, so common is better. 

The separation prevents us from accidentally using unsafe (ie. dereferencing pointers to objects in the target process) probe code in the client.

core/objectidfilterproxymodel.h
pasnox: Done

plugins/quickinspector/quickinspector.cpp
pasnox: Done

pasnox:
Rebased and all comments fixed.

vkrause:
+2

PR #229, opened 2016-06-17
enfranchise tool handling from models.

akreuzkamp:
This change removes ToolModel and ToolSelectionModel. It replaces them by a new ToolManager class, that does the same. On the client side I added the class ClientToolManager to provide the new tool management API. It also offers it's data as a model, but the new model is client sided only. but now you can also access the tools directly without the model. That is needed if we want to explicitly show a specific tool instead of just displaying a list of all available tools, e.g. when integrating into an IDE.

It also contains a commit to add object navigation to the actions inspector. This is simply because I needed it for the unittest :)


vkrause:
I have integrated the QAction changes to master, for the rest I think we have to wait until after the 2.5 branching, otherwise we breakt the QtC 4.0 plug-in, which then requires API that no GammaRay release actually provides yet/anymore. Should be a matter of days only I hope.

We could merge parts immediately though, like the ObjectId move to its own file, if those would be split off into individual commits, in case that is going to simplify merging with 2.5.

krf:
@vkrause: Please have another look. I'll squash the commits as much as I can when merging this branch. I added them as individual commits to keep Anton's initial commit intact.

core/toolmanager.cpp
```diff
@@ -94,7 +94,7 @@ void ToolManager::selectObject(ObjectId id, const QString &toolId)
     }
 }
 
-void GammaRay::ToolManager::selectTool(const QString &toolId)
```
akreuzkamp: Hehe, that was kdevelop-clang ;P

ui/contextmenuextension.cpp
```diff
@@ -118,17 +118,17 @@ void ContextMenuExtension::populateMenu(QMenu *menu)
...
+    auto toolManager = ObjectBroker::object<ToolManagerInterface *>();
+    toolManager->requestToolsForObject(m_id);
 
     // delay adding actions until we know the supported tools
-    connect(probeController, &ProbeControllerInterface::supportedToolsResponse,
+    connect(toolManager, &ToolManagerInterface::toolsForObjectResponse,
             menu, [menu](ObjectId id, const ToolInfos &toolInfos) {
         foreach (const auto &toolInfo, toolInfos) {
-            auto action = menu->addAction(tr("Show in \"%1\" tool").arg(toolInfo.name));
+            auto action = menu->addAction(tr("Show in \"%1\" tool").arg(toolInfo.id));
```
vkrause: This needs to show the user-readable tool name, not the internal id. This however is also broken in 2.5 already, but will likely require a different fix here.

akreuzkamp: yes, I saw this, when rebasing. I decided, that fixing it is out of scope for this change... I'd rather do it in a separate PR.

krf:
Squashed changes & rebased on top of master.

PR #230, opened 2016-06-29
Move the translated name() member ui-wise.

pasnox:
This make it easier to have correctly translated names when using
the QtC GammaRay plugin.

Task-Id: KDEND-74


vkrause:
Integrated into master.

Issue #225, opened 2016-06-09
Geometry preview not always displaying something


tmcguire:
Quick Scenes->Scene Graph->Geometry->Preview is sometimes empty and doesn't show any triangles.

To reproduce, start "gammaray qmlscene test.qml" with the following QML file:

```
import QtQuick 2.0

Text {
  text: "Hello World"
}
```

Then click: Quick Scenes->QQuick/Text->Scene Graph->Geometry Node->Geometry
The preview will be empty.
After clicking on another node and back on the geometry node, the preview appears.

Having this working reliable would be nice, I use this feature to demonstrate that QtQuick uses triangles for rendering in trainings.

Version: Gammaray master (v2.4.1-587-g6f26f6e), Qt 5.5

vkrause:
Yep, seen the same here, the very first selection of a geometry node will not show the geometry.

krf:
@akreuzkamp Fixed by #234?

vkrause:
Yep, #234 fixed this.

PR #234, opened 2016-07-05
Morph indexData bytearray into proper model.

akreuzkamp:
GeometryExtension used to have a custom geometryChanged signal, that
passed on a QByteArray containing the data. That signal needed ugly
decoding on client side and especially wouldn't be fired when first
opening the Geometry tab. Thus now we do the decoding on server side
and feed a proper QAIM from it.

It still doesn't abstract away all of the OpenGL interna (we still use the drawMode on client side to render the wireframe), but I didn't see an easy way to draw highlighted faces without that information.


akreuzkamp:
Fixes #225.

plugins/quickinspector/geometryextension/sggeometryextension.cpp
```diff
@@ -36,12 +36,13 @@
 using namespace GammaRay;
 
 SGGeometryExtension::SGGeometryExtension(PropertyController *controller)
-    : SGGeometryExtensionInterface(controller->objectBaseName() + ".sgGeometry", controller)
-    , PropertyControllerExtension(controller->objectBaseName() + ".sgGeometry")
+    : PropertyControllerExtension(controller->objectBaseName() + ".sgGeometry")
     , m_node(0)
-    , m_model(new SGGeometryModel(this))
+    , m_vertexModel(new SGVertexModel())
+    , m_adjacencyModel(new SGAdjacencyModel())
```
vkrause: While correct from a memory management POV, this will "leak" the two models into the user-visible object tree. If you give them a GammaRay object as parent, they will be filtered out.

vkrause:
Also, this is a protocol change, so this can't go to 2.5, and should increment the protocol version, to avoid having problems with the client trying to talk to an older probe.

vkrause:
Result looks fine on plain rectangles, and indeed shows up immediately now. Seems to fail on rectangles with radius > 0 though.

akreuzkamp:
When exactly does it fail? Here it works with rounded rectangles...

akreuzkamp:
ping

akreuzkamp:
the issue about radius > 0 (which was actually related to changing the count of wires in any way, including just choosing another item) should be fixed now (the client side just missed to react on a modelReset-signal).

PR #235, opened 2016-07-06
CMake: Work-around for AUTOGEN/JSON issue

krf:
Problem: QtAutogen does not regenerate moc file when moc inputs change
See: https://cmake.org/Bug/bug_relationship_graph.php?bug_id=15419

Task-Id: KDEND-44


krf:
Will spend some time fixing this in CMake itself later on (https://cmake.org/Bug/bug_relationship_graph.php?bug_id=15419).

For the time being, we'll need this work-around.

vkrause:
Ugly, but still better than debugging plugin misbehavior due to incomplete rebuilds.

+2

PR #236, opened 2016-07-08
Add per target save/restore states.

pasnox:


krf:
Could you rebase?

Also: Could you squash a few commits? E.g. the "Fix selection update" + revert commit can be removed.

ui/tooluifactory.cpp
```diff
@@ -50,4 +73,37 @@ bool ToolUiFactory::remotingSupported() const
...
+        createExtension<ConnectionsExtensionClient>);
+    PropertyWidget::registerTab<EnumsTab>(QStringLiteral("enums"),
+                                          ObjectInspectorFactory::tr("Enums"),
+                                          PropertyWidgetTabPriority::Exotic - 1);
+    PropertyWidget::registerTab<ClassInfoTab>(QStringLiteral("classInfo"),
+                                              ObjectInspectorFactory::tr("Class Info"),
+                                              PropertyWidgetTabPriority::Exotic - 1);
+    PropertyWidget::registerTab<ApplicationAttributeTab>(
+                QStringLiteral("applicationAttributes"),
+                ObjectInspectorFactory::tr("Attributes"), PropertyWidgetTabPriority::Advanced);
```
vkrause: This looks wrong. What makes the object tool property extensions so special that they need to be in the base class? Ie. why do only these extensions break with a different selection, not those provided by other tools?

Also, why does selection interfere with this at all? initUi() is supposed to be called for all active tools, as soon as they become active. Does this fail, or is that for some reason not enough?

vkrause:
I cherry-picked the first one in this series to 2.5.

ui/tooluifactory.cpp
pasnox: The GammaRay code was written with the assumption that the object inspector is always the default / first tool shown.
And this one init most of the tab for any QObject based inspector.
The other inspectors, only init the new one they need (or custom one), because the other were already loaded / initialized by the object inspector.
Now that we can restore the 'initial' tool at startup, the object one is no more the always first shown, so all the tab it does normally init, are not.
So the restored inspectors is missing anything the object inspector is registering.
That's why this change is needed (unless i'm wrong somewhere, which is possible :))

vkrause: Does that mean that initUi() isn't called for the object inspector as soon as the object tool is activated (not selected by the user, but becomes selectable as the corresponding part on the probe sees the first matching object)? That would be the actual bug then, initUi is separate from the widget ctor exactly for this reason, so we can init property tabs even if the tool isn't actually visible.

pasnox: This change is no longer needed with anton and ready() fix changes.

pasnox: Done

pasnox:
Done

ui/uistatemanager.cpp
```diff
@@ -167,10 +173,32 @@ void UIStateManager::restoreState()
     restoreWindowState();
     restoreSplitterState();
     restoreHeaderState();
+
+    // Allow custom state per end point
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    if (m_customRestoreMethod.isValid())
+#endif
```
vkrause: Does this mean we end up calling invalid methods with Qt4? Might be safer to entirely skip this with Qt4 in that case? (same again below)

ui/uistatemanager.h
```diff
@@ -50,6 +51,15 @@ namespace GammaRay {
  * That mean any not moved splitter handle will not be persisted, and % based sizes
  * will works correctly even on resize.
  * Any QMainwindow is always restored/saved.
+ *
+ * There is now the possibility to implements custom restore/save state per targets.
+ * Just create those 2 public Q_INVOKABLE in the UIStateManager widget:
+ * - Q_INVOKABLE void saveCustomState(QSettings *settings) const;
+ * - Q_INVOKABLE void restoreCustomState(QSettings *settings);
```
vkrause: Wouldn't "Target" instead of "Custom" be the more clear name for this?

ui/uistatemanager.cpp
```diff
@@ -167,10 +173,32 @@ void UIStateManager::restoreState()
...
 
 void UIStateManager::saveState()
 {
+    // Allow custom state per end point
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    if (m_customSaveMethod.isValid())
+#endif
+    {
+        Q_ASSERT(!Endpoint::instance()->label().isEmpty());
+        m_stateSettings->beginGroup(Endpoint::instance()->label());
```
vkrause: label() looks problematic as that can contain the PID IIUC, ie. we would constantly change the settings key here.

ui/uistatemanager.cpp
pasnox: Done

ui/uistatemanager.h
pasnox: Done

ui/uistatemanager.cpp
pasnox: Done

PR #237, opened 2016-07-08
Safer Message reading and writing.

pasnox:
The Message::payload() has been hidden and now only used
by the Message's >> and << operators.
Those operators ensure compile time check on meta types and run time
check on the payload stream status.

### DIFF
--- a/common/message.h
+++ b/common/message.h
@@ -63,10 +63,48 @@ public:
     Protocol::ObjectAddress address() const;
     Protocol::MessageType type() const;
 
-    /** Access to the message payload. This is read-only for received messages
-     *  and write-only for messages to be sent.
+    /** Read value from the payload
+     *  This operator proxy over payload() allow to do:
+     *   - Run time check on the stream status
      */
-    QDataStream &payload() const;
+    template <typename T>
+    GammaRay::Message &operator>>(T &value)
+    {
+        if (Q_UNLIKELY(payload().status() != QDataStream::Ok)) {
+            qWarning("%s: Attempting to read from a non valid stream: status: %i", Q_FUNC_INFO, int(payload().status()));
+        }
+        payload() >> value;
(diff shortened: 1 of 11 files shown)


common/message.h
```diff
@@ -63,10 +63,50 @@ class GAMMARAY_COMMON_EXPORT Message
...
+    /** Write value to the payload.
+     *  This operator proxy over payload() allow to do:
+     *   - Compile time check on the metatype id
+     *   - Run time check on the stream status
+     */
+    template <typename T>
+    GammaRay::Message &operator<<(const T &value)
+    {
+        const int id = qMetaTypeId<T>();
+        const char *name = QMetaType::typeName(id);
```
vkrause: Maybe move those lines out of the good case path?

common/message.h
```diff
@@ -63,10 +63,50 @@ class GAMMARAY_COMMON_EXPORT Message
...
+        if (payload().status() != QDataStream::Ok)
```
vkrause: Candidate for Q_UNLIKELY, these methods are on a hot path I think.

krf:
Rest looks good to me, :+1:

vkrause:
Tested this,  it seems too strict actually. It fails on containers with older Qt versions (prior to those being auto-registered as meta types), yet that has absolutely no impact on QDataStream (de)serialization. What's the reason behind enforcing meta type registration?

common/message.h
pasnox: Done

common/message.h
pasnox: Done

pasnox:
I will try to check builds errors with oldest Qt.
Enforcing the metatype registration make sure we can send queued events for the used types.
I lost some time finding that one of the root of a broken message comes from an unseen deserialization error hidden in code which can also not be ours (yeah, you QVariantMap). Some QVariant were not (de)serializable and silently fails, just the stream status was updated, but we were not doing any check one that.

PR #238, opened 2016-07-14
Style fixes

krf:


ui/mainwindow.cpp
```diff
@@ -115,31 +151,23 @@ MainWindow::MainWindow(QWidget *parent)
     , ui(new Ui::MainWindow)
     , m_stateManager(this)
 {
+
+    const auto styleOverride = gammarayStyleOverride();
+    if (styleOverride) {
+        applyStyle(styleOverride);
```
vkrause: The applyStyle() hack is only needed for in-process mode, for the normal client qApp->setStyle() is a lot easier and more robust.

vkrause:
+2

You can already test different styles by launching the client manually and using the -style option, but I'm fine with the env var solution as well.

ui/mainwindow.cpp
krf: Yeah, I wanted to keep this in line with what's done for the out-of-process mode. Plus the style changes are only local to the main window.

krf:
Yep, I'm aware of that. It's way easier to be able to control this within an environment variable, though.  Working with `gammaray-client` is extra work.

PR #240, opened 2016-07-28
Fix compilation on ARM when building against Qt 4

xhaakon:
error: no matching function for call to 'qBound(qreal, double, qreal)'
  note: deduced conflicting types for parameter 'const T' ('float' and 'double')

It was happening because qreal 'is a typedef for float for performance reasons'
on ARM in Qt 4.

### DIFF
--- a/client/messagestatisticsmodel.cpp
+++ b/client/messagestatisticsmodel.cpp
@@ -167,8 +167,8 @@ static const int GRADIENT_SCALE_FACTOR = 4;
 
 static QColor colorForRatio(double ratio)
 {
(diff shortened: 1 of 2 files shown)


winterz:
looks like we'll need the same type of fix in client/messagestatisticsmodel.cpp too?
could you look please?

I suppose we could cast all 3 arguments as qreal's but I like your solution better.

xhaakon:
> looks like we'll need the same type of fix in client/messagestatisticsmodel.cpp too?

Didn't catch that one because for Debian we build only the Qt 4 probes, but in messagestatisticsmodel.cpp it's the very same situation, so I'd expect compilation would fail there as well.

I'll update the commit to include fixes in both places.

PR #242, opened 2016-08-12
Wayland compositor probe

giucam:
This PR adds a probe plugin that provides information about a QtWayland compositor. It shows the list of connected clients, their resources, and various information about these resources, based on their type (such as type, version, surface title, surface content, ...)
It also has a view logging the protocol flow and a timeline, though that is still very rough.

It needs a recent QtWayland and a very recent libwayland (that is, a few hours old now).
wayland 1.12 will be released in a couple of weeks, so then we'll be able to add a requirement in CMakeLists.txt.


vkrause:
Pushed this branch to the main repo, so see what the CI thinks about it.

vkrause:
Integrated into master. Thanks!

PR #243, opened 2016-08-12
Fix Qt Core and ABI detection on embedded Linux systems without ldd

milianw:
This patch adds a fallback which gets run when ldd could not be
started. In that case, we instead launch the target debuggee binary
directly, but set the LD_TRACE_LOADED_OBJECTS=1 environment
variable. This makes ld.so behave like ldd, and instead of executing
the debuggee it will output the desired information about dynamic
dependencies.

### DIFF
--- a/launcher/probeabidetector_elf.cpp
+++ b/launcher/probeabidetector_elf.cpp
@@ -34,6 +34,7 @@
 #include <QDebug>
 #include <QFile>
 #include <QProcess>
+#include <QProcessEnvironment>
 #include <QString>
 #include <QStringList>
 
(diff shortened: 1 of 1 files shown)


milianw:
Before:

```
gammaray ./trivial    
GammaRay server listening on: tcp://192.168.150.13:11732
```

After:
```
# gammaray ./trivial 
Probe ABI:qt5_6-aarch64
Detected ABI qt5_6-aarch64, using ABI qt5_6-aarch64.
QIODevice::read (QProcess): device not open
GammaRay server listening on: tcp://192.168.150.13:11732
```

vkrause:
+2

PR #244, opened 2016-08-12
Check whether we can read a line before doing do

milianw:
Fixes a runtime warning saying:

QIODevice::read (QProcess): device not open

### DIFF
--- a/launcher/probeabidetector_elf.cpp
+++ b/launcher/probeabidetector_elf.cpp
@@ -69,7 +69,7 @@ static QString qtCoreFromLdd(const QString &path, bool fallback = false)
     }
     proc.waitForFinished();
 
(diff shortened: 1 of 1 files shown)


winterz:
+1
should also be backported to the 2.5 branch

milianw:
should I push it to that branch and then merge it up, or cherry-pick it from master into 2.5?

vkrause:
Push to 2.5 and merge (or just wait for the next merge, there is usually one per day anyway).

vkrause:
I've cherry-picked this change to master now. Turns out the next release has to be 2.6 anyway, due to the upcoming QtC 4.1 release, so backporting to 2.5 is no longer relevant. Thanks!

PR #245, opened 2016-08-16
Fix detecting top level windows

pasnox:


plugins/guisupport/guisupport.cpp
```diff
@@ -347,6 +350,25 @@ void GuiSupport::registerVariantHandler()
...
+    }
+}
+
+void GuiSupport::objectCreated(QObject *object)
+{
+    if (!object)
+        return;
+
+    if (qobject_cast<QGuiApplication *>(object))
+        discoverObjects();
```
vkrause: As this is very high-volume, we might want to disconnect objectCreated here again. Also, both null checks here and in discoverObjects are unnecessary, Probe::objectCreated guarantees a valid pointer here, and qApp above is verified by the qobject_cast down here already.

+2 in any case

PR #246, opened 2016-08-16
Fix selecting a wayland resource

giucam:
For some reason using uint32_t worked correctly with the wayland QPA, but with
the xcb QPA it didn't, and QMetaObject would complain:

QMetaObject::invokeMethod: No such method GammaRay::WlCompositorInspector::setSelectedResource(uint)
Candidates are:
    setSelectedResource(uint32_t)
    setSelectedResource(uint32_t)

Using uint instead works always.


vkrause:
Looks like missing meta type alias registrations for uint32_t, your change should indeed make this more robust. +2

PR #250, opened 2016-09-26
Fix changing the QSG visualize modes

giucam:
When using the threaded renderer we need to make sure to recreate it
on the correct thread.


akreuzkamp:
You're my hero of the day! Thank you! :)

PR #252, opened 2016-10-02
Use disabled foreground palette for hidden widgets in widget model.

milianw:
This feature was broken in the out-of-process client/server mode.
I noticed this recently while demoing the tool for a customer on
a widget application that contained multiple top-level widgets.
Contrary to my first expectations, only one of them was visible
at any given time. This patch makes this distinction clear by
applying the same mechanism used in the quick inspector also for
widgets, namely introducing a client-side proxy model that applies
the styling filter. The server-side tree model is taught a new
VisibleRole.


vkrause:
+2, thanks!

PR #257, opened 2016-10-14
Make ClientToolManager persistent.

akreuzkamp:
This commit allows a ClientToolManager instance to survive a disconnect
and reset itself, so when reconnecting the client to another server,
there is no need to delete and reconstruct the ClientToolManager.

To achieve this, this commit moves the connectionEstablished signal from
Client to Endpoint (for linking reasons). I also added an emit to
Server, just to not have misleading API.

### DIFF
--- a/client/client.h
+++ b/client/client.h
@@ -66,8 +66,6 @@ public:
     void unregisterMessageHandler(Protocol::ObjectAddress objectAddress) Q_DECL_OVERRIDE;
 
 signals:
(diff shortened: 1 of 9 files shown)


pasnox:
Please rework/rebase as https://github.com/KDAB/GammaRay/pull/236 introduced conflicts.

akreuzkamp:
Hey, I'd need this merged as soon as possible, to get my qtc-integration-patches in.
Thanks :)

PR #258, opened 2016-10-17
Fix checking if a quick item is partially or completely out of view

giucam:
In some cases the QtQuick plugin would say that an item is out of view
even if it is only partially so:

```
import QtQuick 2.2
Item {
        width: 400
        height: 400

        Rectangle {
                color: "green"
                x: 300
                y: 300
                width: 20
                height: 200
                scale: -1
        }
}
```

Distinguish between partialy and completely out of view and make sure
to report that correctly.

### DIFF
--- a/plugins/quickinspector/quickitemmodel.cpp
+++ b/plugins/quickinspector/quickitemmodel.cpp
@@ -353,21 +353,30 @@ void QuickItemModel::updateItemFlags(QQuickItem *item)
 {
     QQuickItem *ancestor = item->parentItem();
     bool outOfView = false;
+    bool partiallyOutOfView = false;
+
+    auto rect = item->mapRectToScene(QRectF(0, 0, item->width(), item->height()));
+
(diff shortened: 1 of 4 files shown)


plugins/quickinspector/quickitemmodel.cpp
```diff
@@ -351,23 +351,21 @@ void QuickItemModel::updateItem(QQuickItem *item, int role)
...
-                || pos.x() > ancestor->width() || pos.y() > ancestor->height())) {
-            outOfView = true;
-            break;
-        }
-        ancestor = ancestor->parentItem();
-    }
+    bool partiallyOutOfView = false;
+
+    auto rect = item->mapRectToScene(QRectF(0, 0, item->width(), item->height()));
+    auto sceneRect = QRectF(0, 0, m_window->width(), m_window->height());
```
giucam: Fixed.

PR #260, opened 2016-10-22
Implement touch events redirection from the QtQuick preview widget

giucam:
This is based on https://github.com/KDAB/GammaRay/tree/master-KDEND-68 
It now works properly, as far as i can see.

### DIFF
--- a/common/remoteviewinterface.cpp
+++ b/common/remoteviewinterface.cpp
@@ -37,15 +37,112 @@ QT_BEGIN_NAMESPACE
 GAMMARAY_ENUM_STREAM_OPERATORS(RemoteViewInterface::RequestMode)
 QT_END_NAMESPACE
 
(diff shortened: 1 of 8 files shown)


common/remoteviewinterface.cpp
```diff
@@ -37,15 +37,112 @@ QT_BEGIN_NAMESPACE
...
+    int f;
+    s >> f;
+    flags = QTouchEvent::TouchPoint::InfoFlags(f);
+    return s;
+}
+
+QDataStream &operator<<(QDataStream &s, const QList<QTouchEvent::TouchPoint> &points)
+{
+    s << points.count();
+    for (const auto &p: points) {
```
akreuzkamp: we still support MSVC 2010, which doesn't support range-based for... The CI should have caught that, right?

krf:
@giucam Anything left here?

giucam:
No, i think this is good to go.

PR #261, opened 2016-10-22
IOS Toolchain for static probe

oberluz:
Enable building an IOS  static probe library

### DIFF
--- a/3rdparty/kde/CMakeLists.txt
+++ b/3rdparty/kde/CMakeLists.txt
@@ -4,7 +4,7 @@ set(gammaray_kitemmodels_srcs
   krecursivefilterproxymodel.cpp
 )
 
-add_library(gammaray_kitemmodels SHARED ${gammaray_kitemmodels_srcs})
+add_library(gammaray_kitemmodels ${GAMMARAY_LIBRARY_TYPE} ${gammaray_kitemmodels_srcs})
 
 if(Qt5Core_FOUND)
   target_link_libraries(gammaray_kitemmodels LINK_PUBLIC Qt5::Core)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,12 +34,7 @@
 
 project(GammaRay CXX C)
 
-if (MSVC)
-  cmake_minimum_required(VERSION 3.1.0) # b/c of TARGET_PDB_FILE_DIR generator expression
-else()
-  cmake_minimum_required(VERSION 2.8.12) # b/c of ECMGeneratePriFile.cmake
-endif()
-
+cmake_minimum_required(VERSION 2.8.12) # b/c of ECMGeneratePriFile.cmake
 cmake_policy(SET CMP0020 NEW)
 if(POLICY CMP0043)
   cmake_policy(SET CMP0043 NEW)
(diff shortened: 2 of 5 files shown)


3rdparty/kde/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set(gammaray_kitemmodels_srcs
   krecursivefilterproxymodel.cpp
 )
 
-add_library(gammaray_kitemmodels SHARED ${gammaray_kitemmodels_srcs})
+add_library(gammaray_kitemmodels ${GAMMARAY_LIBRARY_TYPE} ${gammaray_kitemmodels_srcs})
```
pasnox: The proper way would have been to introduce a 'real' static build option to the entire gammaray, especially at least for the probe and it's actually shared only plugins - grep for gammaray_add_plugin to find them.

CMakeLists.txt
```diff
@@ -233,8 +233,8 @@ endif()
 # linker flags
 if(NOT ECM_ENABLE_SANITIZERS AND (CMAKE_SYSTEM_NAME MATCHES Linux OR CMAKE_SYSTEM_NAME STREQUAL GNU))
   if(CMAKE_COMPILER_IS_GNUCXX OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    set(CMAKE_SHARED_LINKER_FLAGS "-Wl,--fatal-warnings -Wl,--no-undefined -lc ${CMAKE_SHARED_LINKER_FLAGS}")
-    set(CMAKE_MODULE_LINKER_FLAGS "-Wl,--fatal-warnings -Wl,--no-undefined -lc ${CMAKE_MODULE_LINKER_FLAGS}")
+    set(CMAKE_SHARED_LINKER_FLAGS "-lc ${CMAKE_SHARED_LINKER_FLAGS}")
```
pasnox: why removing the flags ?

CMakeLists.txt
```diff
@@ -233,8 +233,8 @@ endif()
...
+    set(CMAKE_MODULE_LINKER_FLAGS "-lc ${CMAKE_MODULE_LINKER_FLAGS}")
```
pasnox: Same

Install.txt
```diff
@@ -70,6 +70,25 @@ or
 $ adb forward --remove-all
 to remove all forwards
 
+Building Gammaray probe for IOS (OSX "El Capitan" / Qt 5.7 / Xcode 8 / iphone SDK 10.0):
+Make sure you have gdb and qmake in your path
+% export QT_DIR=<QT HOME> <--- set to QT's home
+% mkdir build
+% cd build
+% cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-IOS.cmake -DCMAKE_PREFIX_PATH=${QT_DIR}/5.7/ios -DGAMMARAY_STATIC_PROBE=TRUE -DCMAKE_INSTALL_PREFIX=$SYSROOT/usr/ -DGAMMARAY_BUILD_UI=OFF -G Xcode ..
+% /usr/bin/xcodebuild -project GammaRay.xcodeproj build -target gammaray_probe -configuration Release
```
pasnox: If requested changes are done, then a simple: make is needed.

cmake/Toolchain-IOS.cmake
```diff
@@ -0,0 +1,66 @@
+# Basic cmake toolchain file for IOS
```
pasnox: I would have simply duplicate the OSX one and change i686 to arm (or armv7).
If any errors, report them to us.

oberluz: I tried that + unix makefiles as the backend but I'm afraid I could not get it to generate arm code, only x86. On my machine at least it's not as simple as just duplicating the OSX toolchain and changing the architecture. It could be my setup is broken for gcc. Feel free to decline this PR as it doesn't really fit your requirements.

PR #263, opened 2016-10-28
Fix missing code navigation support

pasnox:
The issue appear when attaching a qml application
and right clicking an object in any QObject based inspector.

Task-Id: KDEND-41

### DIFF
--- a/plugins/quickinspector/quickinspector.cpp
+++ b/plugins/quickinspector/quickinspector.cpp
@@ -262,6 +262,11 @@ QuickInspector::QuickInspector(ProbeInterface *probe, QObject *parent)
     filterProxy->addRole(ObjectModel::ObjectIdRole);
     probe->registerModel(QStringLiteral("com.kdab.GammaRay.QuickItemModel"), filterProxy);
 
+    if (m_probe->needsObjectDiscovery()) {
+        connect(m_probe->probe(), SIGNAL(objectCreated(QObject*)),
+                SLOT(objectCreated(QObject*)));
+    }
+
     connect(probe->probe(), SIGNAL(objectCreated(QObject*)),
             m_itemModel, SLOT(objectAdded(QObject*)));
     connect(probe->probe(), SIGNAL(objectDestroyed(QObject*)),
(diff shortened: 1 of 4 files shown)


plugins/quickinspector/quickinspector.cpp
```diff
@@ -387,6 +389,28 @@ void QuickInspector::objectSelected(void *object, const QString &typeName)
         selectSGNode(reinterpret_cast<QSGNode *>(object));
 }
 
+void QuickInspector::objectCreated(QObject *object)
+{
+    if (!object)
```
vkrause: unnecessary check, Probe guarantees object to not be null, and the next if() tests the same implicitly as well

plugins/quickinspector/quickinspector.cpp
```diff
@@ -387,6 +389,28 @@ void QuickInspector::objectSelected(void *object, const QString &typeName)
...
+    if (QQuickWindow *window = qobject_cast<QQuickWindow *>(object)) {
+        if (QQuickView *view = qobject_cast<QQuickView *>(object)) {
+            m_probe->discoverObject(view->engine());
+        }
+        else {
+            QQmlContext *context = QQmlEngine::contextForObject(window);
+            QQmlEngine *engine = context ? context->engine() : Q_NULLPTR;
+
+            if (!engine) {
+                engine = qmlEngine(window->contentItem()->childItems().value(0));
```
vkrause: Does this need a check for childItems() being empty?

plugins/quickinspector/quickinspector.cpp
```diff
@@ -262,6 +262,8 @@ QuickInspector::QuickInspector(ProbeInterface *probe, QObject *parent)
     filterProxy->addRole(ObjectModel::ObjectIdRole);
     probe->registerModel(QStringLiteral("com.kdab.GammaRay.QuickItemModel"), filterProxy);
 
+    connect(m_probe->probe(), SIGNAL(objectCreated(QObject*)),
```
vkrause: This can be conditional on m_probe->needsObjectDiscovery(), to avoid the extra performance cost when we are not attaching at runtime.

plugins/quickinspector/quickinspector.cpp
pasnox: no, value() is compatible with out of bounds index and would simply return null.

plugins/quickinspector/quickinspector.cpp
pasnox: Done

plugins/quickinspector/quickinspector.cpp
pasnox: Done

PR #266, opened 2016-11-02
Language edits and correct typos

veshivas:
Also updated some section titles to use title case instead
of sentence case.

### DIFF
--- a/docs/manual/gammaray-action-inspector.qdoc
+++ b/docs/manual/gammaray-action-inspector.qdoc
@@ -26,12 +26,12 @@
     \section1 Overview
 
     The action inspector shows all QAction instances that exist in the target application. Information about
-    the display name, checkable states and shortcuts and shortcut priorities are displayed.
+    the display name, checkable states, and shortcuts with their priorities are displayed.
 
-    \image gammaray-action-inspector.png
+    \borderedimage gammaray-action-inspector.png
 
-    The action inspector will warn about shortcut conflicts (that is, two QAction instances using the same
+    The action inspector warns about shortcut conflicts (that is, two QAction instances using the same
     shortcut sequence and thus making it ambiguous) with a warning icon in its shortcut column.
 
-    Actions can also be activated remotely via GammaRay, by double-clicking the corresponding entry.
+    Actions can also be activated remotely by double-clicking the corresponding entry.
 */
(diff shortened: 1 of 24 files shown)


docs/manual/gammaray-qresource-browser.qdoc
```diff
@@ -27,16 +27,16 @@
 
     The resource browser view allows to inspect the content of assets compiled into the application binary using the QResource system.
     The tree view on the left side shows the file and folder hierarchy, the content view on the right allows to inspect the file content
-    if it is a supported text or image format.
+    if it is a supported text or borderedimage format.
```
vkrause: search/replace mistake?

docs/manual/gammaray-action-inspector.qdoc
```diff
@@ -26,12 +26,12 @@
     \section1 Overview
 
     The action inspector shows all QAction instances that exist in the target application. Information about
-    the display name, checkable states and shortcuts and shortcut priorities are displayed.
+    the display name, checkable states, and shortcuts with their priorities are displayed.
 
-    \image gammaray-action-inspector.png
+    \borderedborderedimage gammaray-action-inspector.png
```
vkrause: double search/replace? occurs in a few places below as well

vkrause:
Tried to integrate this now, but I'm getting "warning: Unknown command '\borderedimage'" for each instance of \borderedimage (with Qt 5.7), when building offline docs, and no images in the output anymore.

veshivas:
The "\borderedimage" macro was added in 9f45d2ab03a643348c6f0c464aa48755b301498d. Looks like your 5.7 build is missing that. The quickest way to fix this is to patch the macros.qdocconf, which you'll find under the QT_INSTALL_DOCS path. Run 'qmake -query' to know this path, and append "macro.borderedimage       = "\\div {class=\"border\"} \\image \1\n\\enddiv"" to macros.qdocconf.

vkrause:
Same problem with Qt 5.8 for both offline and online docs, and I've verified the macro is defined there. Weird. Anyway, as GammaRay supports Qt versions all the way back to 4.8 I need to define this macro locally anyway for compatibility, and then it seems to work fine. I'll integrate your patch as is then, thanks!

Issue #105, opened 2014-12-19
Doesn't highlight preview when clicking on sublayouts


iamsergio:
When going to widgets-> then clicking on nested layouts, the highlight remains on the top layout.

```
#include <QApplication>
#include <QDebug>
#include <QWidget>
#include <QColorDialog>

int main(int a, char**b)
{
    QApplication app(a, b);
    QColorDialog dialog;
    dialog.exec();
    return app.exec();
}
```

iamsergio:
is this still an issue ?

krf:
Yes, PR is still not merged.

krf:
PR got merged

PR #268, opened 2016-11-03
Properly draw overlay for layouts

danvratil:
Fixes issue #105

### DIFF
--- a/plugins/widgetinspector/widgetinspectorserver.cpp
+++ b/plugins/widgetinspector/widgetinspectorserver.cpp
@@ -170,19 +170,20 @@ void WidgetInspectorServer::widgetSelected(const QItemSelection &selection)
     }
     const QModelIndex index = selection.first().topLeft();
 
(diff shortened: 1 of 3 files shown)


plugins/widgetinspector/widgetinspectorserver.cpp
```diff
@@ -195,15 +196,18 @@ void WidgetInspectorServer::widgetSelected(const QItemSelection &selection)
     if (m_selectedWidget
         && (qobject_cast<QDesktopWidget *>(m_selectedWidget)
             || m_selectedWidget->inherits("QDesktopScreenWidget"))) {
-        m_overlayWidget->placeOn(0);
+        m_overlayWidget->placeOn({});
```
vkrause: {} will probably be rejected by MSVC 2010/2012 (which are only in the nightly builds, so you need to trigger a build there manually to verify before integration).

PR #271, opened 2016-11-15
Merge Widget 3D Inspector

danvratil:
Widget 3D Inspector visualizes the widget hierarchy by exploding it along the Z axis. Users can rotate the window in a 3D space, zoom in and out (using the + and - keys) and use mouse wheel to collapse/expand the tree.

gammaray1

There's also the "Wireframe" mode, which draws a thin wireframe around each widget to better visualize multiple adjacent widgets, and a "Horizontals" mode, which draws horizontals from widget to its parent, better visualizing the exact position of the child in the parent.

gammaray2

When hovering mouse cursor over a widget the panel in top right corner shows detailed informations about the widget. Right-clicking a widget will pop up a context menu to show the widget in the other inspectors).

The server-side plugin is compiled for all versions of Qt 5, the client-side viewer is only compiled when Qt3D >= 5.7 is available (due to API changes and/or bugs in previous versions of Qt3D). On the server-side we provide `com.kdab.GammaRay.Widget3DModel` built on top of `ObjectModel` which introduces a bunch of new roles, including textures of the widgets. On the client side we use a little hack to circumvent the `RemoteModel` being lazily populated (since we need to visualize the whole thing at once) and a bunch of proxies. We find top-level windows in the tree and put them into the combobox on top. Based on the selected window we filter out all subtrees that don't belong to that window (including windows which are children (of children) of the current window) and flatten the tree so we can feed it to Qt3D `NodeInstantiator`. In Qt3D we simply create a new cube for each widget. Maybe someone with actual Qt3D/OpenGL knowledge could look at my shaders and anti-aliasing (or rather lack of it, although it does not look too bad).

Any ideas for more features/visualizations are welcomed, I consider this PR just as an initial review and a call for feedback. We can potentially merge it without enabling this in the UI for now but to have it available for showcasing GammaRay, there's certainly lots of potential for improvements. 

Millian (and Volker too IIRC) suggested it would be cool to have the same 3D visualization for QML scenes, but maybe with the exception of the `Widget3DSubtreeModel` there's not much code that could be shared, so I'd consider that a standalone feature to look into once this is done :-)


plugins/widgetinspector/widgetinspectorserver.cpp
```diff
@@ -133,6 +136,12 @@ WidgetInspectorServer::WidgetInspectorServer(ProbeInterface *probe, QObject *par
 
     probe->registerModel(QStringLiteral("com.kdab.GammaRay.WidgetTree"), widgetSearchProxy);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
vkrause: Shouldn't this be >= 5.7, or even WITH_WIDGET3D?

plugins/widgetinspector/widgetinspectorwidget.cpp
```diff
@@ -131,6 +137,7 @@ WidgetInspectorWidget::WidgetInspectorWidget(QWidget *parent)
     m_stateManager.setDefaultSizes(ui->previewSplitter, UISizeVector() << "50%" << "50%");
 
     connect(ui->widgetPropertyWidget, SIGNAL(tabsUpdated()), this, SLOT(propertyWidgetTabsChanged()));
+    connect(ui->tabWidget, SIGNAL(currentChanged(int)), this, SLOT(onTabChanged(int)));
 }
```
vkrause: Could we maybe hide the tab (and tab bar) for !WITH_WIDGET3D?

plugins/widgetinspector/assets/qml/main.qml
```diff
@@ -0,0 +1,193 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+import QtQuick 2.5 as QQ2
+import QtQml 2.2 as Qml2
+import QtQuick.Controls 1.2 as QQControls
```
danvratil: Added a simple runtime check to `WidgetInspectorWidget`

PR #273, opened 2016-11-22
Fix reentrancy issue

pasnox:
There was a recurrent crash when disconnectFromHost was called too much
in a short time.
Also the invalidated socket was still connected and produced responses
for readyRead / disconnected.

### DIFF
--- a/client/client.cpp
+++ b/client/client.cpp
@@ -89,8 +89,7 @@ void Client::connectToHost(const QUrl &url, int tryAgain)
 
     connect(m_clientDevice, SIGNAL(connected()), this, SLOT(socketConnected()));
     connect(m_clientDevice, SIGNAL(transientError()), this, SIGNAL(transientConnectionError()));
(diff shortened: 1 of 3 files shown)


client/client.cpp
```diff
@@ -114,7 +114,7 @@ void Client::socketConnected()
 void Client::resetClientDevice()
 {
     if (m_clientDevice) {
-        m_clientDevice->deleteLater();
+        QMetaObject::invokeMethod(m_clientDevice, "deleteLater", Qt::QueuedConnection);
```
vkrause: deleteLater() is already queued through the event loop, are you sure this change is ncessary?

pasnox: Indeed it is 'normally' already queued, though it still make me troubles without invoking it that way, I did not success to find why.
Probably The QtC plugin still making too much call here in a short, but with my latest patches in the plugin, I can not see reentrancy again, though resetClientDevice() is still tried to be called twice in a short (probably due to the different connection pointed to resetClientDevice()).

PR #277, opened 2016-12-02
IVI property model: allow to reset the heuristically determined "not really writable" flag

ahartmetz:
...and display it as half checked.


ahartmetz:
Maybe check if the UI isn't too confusing. The code isn't that interesting.

PR #280, opened 2016-12-06
Fixed divide by zero crash in BufferModel

vpicaver:
This patch prevents the BufferModel from crashing
while calculating m_rowSize in BufferModel::updateAttribute

### DIFF
--- a/plugins/qt3dinspector/geometryextension/buffermodel.cpp
+++ b/plugins/qt3dinspector/geometryextension/buffermodel.cpp
@@ -83,16 +83,18 @@ void BufferModel::updateAttributes()
 
 void BufferModel::updateAttribute(const GammaRay::Qt3DGeometryAttributeData &attrData)
 {
-    m_rowSize = m_buffer.size() / attrData.count;
-    for (uint i = 0; i < std::max(attrData.vertexSize, 1u); ++i) {
-        ColumnData col;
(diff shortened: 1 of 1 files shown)


plugins/qt3dinspector/geometryextension/buffermodel.cpp
```diff
@@ -83,16 +83,18 @@ void BufferModel::updateAttributes()
 
 void BufferModel::updateAttribute(const GammaRay::Qt3DGeometryAttributeData &attrData)
 {
-    m_rowSize = m_buffer.size() / attrData.count;
```
krf: @vkrause: The update of `m_rowSize` in this method looks odd. Is that correct?

In case `updateAttribute` is called multiple times in `updateAttributes` (is that possible?), the value is overwritten. What am I missing?

vpicaver: Yea, I'm not sure. I only added the if statement around the original code. The if statement prevented a crash when loading buffers that had no data, I think.

vkrause:
Does b1c85478bf6e2e824c49a285aa4578e7c507e8ed fix this problem for you? It filters out the invalid/empty attributes at the source already.

vpicaver:
Yea, b1c8547 fixes the issue.

PR #282, opened 2016-12-10
statemachineviewer: Add support for scxml log

jpetersen:
Display log messages created via scxml <log> in the statemachineviewer log.


krf:
LGTM

PR #284, opened 2016-12-14
Add server side decorations through an overlay item

pasnox:
Any related Qt issues in code has been fixed in those CR, though we workaround them and does not depend them:
https://codereview.qt-project.org/#/c/179790/
https://codereview.qt-project.org/#/c/179789/
https://codereview.qt-project.org/#/c/179863/


krf:
@pasnox This patch seems to cause quickinspectortest to fail on all platforms now. Could you check?

krf:
Reported at #298, please comment there.

PR #286, opened 2016-12-16
Remove QMetaObjects from the meta objects model when no object is alive

giucam:
This fixes a use-after-free with the qtscxml's invoke-dynamic example.
However it is not a complete fix, because it won't work correctly if
the QMetaObject is destroyed before all its objects are. There is no
way to detect that occurrence, unfortunately.

Task-Id: KDEND-128


vkrause:
Removing the entry means you wont be able to see instance counts for short-lived objects anymore, even if they are created in large quantities and have static meta objects, right? Thinking of e.g. QNetworkReply.

giucam:
Right, but i don't see another solution, bar adding more api around QMetaObject and friends, and making the qtsxcml example use it.

giucam:
I pushed a new version which instead of removing the QMetaObjects it now instead cache the className so that can be fetched still, and keeps it in the model.

giucam:
Ping

vkrause:
Integrated into master.

PR #288, opened 2016-12-19
Ui: Prevent toolSelector-scrollbar from looking ugly

chsterz:
StyleSheet needed to be enhanced with a selector
(QWidget#toolSelector{...}) in order to prevent inheritance.
This apparently affected the Scrollbar.
gamma_why

### DIFF
--- a/ui/mainwindow.ui
+++ b/ui/mainwindow.ui
@@ -33,10 +33,12 @@
        <enum>Qt::CustomContextMenu</enum>
       </property>
       <property name="styleSheet">
-       <string notr="true">background-image: url(:/gammaray/kdab-gammaray-logo.png);
+       <string notr="true">QWidget#toolSelector{
+background-image: url(:/gammaray/kdab-gammaray-logo.png);
 background-repeat: no-repeat;
 background-attachment:fixed;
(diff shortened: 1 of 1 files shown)


ui/mainwindow.ui
```diff
@@ -89,7 +91,7 @@ background-position:bottom right;
      <x>0</x>
      <y>0</y>
      <width>829</width>
-     <height>27</height>
+     <height>39</height>
```
milianw: unrelated?

chsterz: I used the designer to change the stylesheet.
Was wondering myself about this.
So what I did was resetting the file and redo the changes to the stylesheet.
(I carefully looked not to accidentally move something in the designer) 
But again, this change co-ocurred after changing the style, so I thought if the designer thinks this is relevant to the change, I just commit that too.

milianw: Designer really likes to mess with dimensions when you edit something unrelated. Personally, I usually discard these using git after the fact. But I'm not forcing you to do it, just wanted to point it out - you already have my :+1:  :)

krf:
:+1:, merging

PR #289, opened 2016-12-21
Add inbound object navigation for the state machine editor

ahartmetz:
Task-Id: KDEND-108


krf:
Note: This does not seem to work as intended for me:

Howto reproduce:
- Go to Objects tool
- Find QState instance
- Right-click, select 'Show in "State Machines" tool'

Actual:
- GammaRay switches to SM tool but does not select the state in the state model

Expected:
- GammaRay switches to SM tool but & does select the state in the state model

Please check.

vkrause:
Looks good. We should also add support for the other supported types (scxml machines, transitions), but that could be a separate task.

vkrause:
Fixed and integrated.

PR #291, opened 2016-12-23
Exit correctly when the launched process terminates quickly

giucam:
Task-Id: KDEND-124

### DIFF
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -278,8 +278,7 @@ void Launcher::restartTimer()
 
 void Launcher::checkDone()
 {
-    if (d->state == Complete
-        || (d->options.uiMode() == LaunchOptions::InProcessUi && d->state == InjectorFinished)) {
+    if (d->state == InjectorFinished) {
         emit finished();
     } else if ((d->state & InjectorFailed) != 0) {
         d->client.terminate();
--- a/tests/earlyexittest.cpp
+++ b/tests/earlyexittest.cpp
@@ -1,0 +1,75 @@
+/*
+  earlyexittest.cpp
+
+  This file is part of GammaRay, the Qt application inspection and
+  manipulation tool.
+
+  Copyright (C) 2016 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Giulio Camuffo <giulio.camuffo@kdab.com>
+
+  Licensees holding valid commercial KDAB GammaRay licenses may use this file in
+  accordance with GammaRay Commercial License Agreement provided with the Software.
+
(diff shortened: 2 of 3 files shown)


tests/earlyexittest.cpp
```diff
@@ -0,0 +1,71 @@
...
+        // setting the probe is not strictly needed but we silence a runtime warning this way
+        options.setProbeABI(ProbeFinder::listProbeABIs().first());
+        options.setLaunchArguments(QStringList() << "sleep" << "1");
+        Launcher launcher(options);
+
+        QSignalSpy spy(&launcher, SIGNAL(finished()));
+
+        Q_ASSERT(launcher.start());
+
+        int loops = 0;
```
krf: Isn't that what `QSignalSpy::wait` does?

tests/earlyexittest.cpp
```diff
@@ -0,0 +1,71 @@
+/*
+  earlyexittest.cpp
+
+  This file is part of GammaRay, the Qt application inspection and
+  manipulation tool.
+
+  Copyright (C) 2016 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Volker Krause <volker.krause@kdab.com>
```
krf: You're the author ;)

launcher/launcher.cpp
```diff
@@ -278,7 +278,7 @@ void Launcher::restartTimer()
 
 void Launcher::checkDone()
 {
-    if (d->state == Complete
+    if (d->state == InjectorFinished
```
krf: This expression now does some redundant checks: Right now it's `a || (b && a)` => Could be simplified to just `a`. Not sure if that's intended.

I'm not sure if this change is fine, that's probably something for @vkrause.

tests/earlyexittest.cpp
giucam: Right :)

tests/earlyexittest.cpp
giucam: Yes but it was added in 5.0, so for 4.8 we'd need this trick anyway.

launcher/launcher.cpp
giucam: Right... no, it wasn't intended.

tests/earlyexittest.cpp
krf: Ah, got it. I'd still use `QSignalSpy::wait` conditionally there so we can easily detect the custom code as soon as Qt5 will be a required dependency.

giucam:
@krf Does the approval mean you are ok with this even without volker's ok or did you get it through other means?

PR #295, opened 2016-12-28
Meta Objects: Don't warn about internal methods

krf:
With the introduction of private _q_createJSWrapper() in QQuickItem the
entire class hierarchy gets flagged in the meta object browser for having
slots that aren't dynamically invokable. The warning is of dubious use
for private methods though, as invoking those dynamically is far less
common/needed.

So let's just warn about issues in the public interface.

### DIFF
--- a/core/qmetaobjectvalidator.cpp
+++ b/core/qmetaobjectvalidator.cpp
@@ -61,9 +61,12 @@ QMetaObjectValidatorResult::Results QMetaObjectValidator::checkMethod(const QMet
 
     // check for parameters with unknown type
 #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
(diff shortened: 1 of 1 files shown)


krf:
We have similar issues with: `void map{From,To}{Global,Item}(QQmlV4Function*)`

We *could* filter them out by checking whether any of the types of the method parameters starts with "QQmlV4". This would mean more QML-specific code/checks in `QMetaObjectValidator` though.

Opinions?

giucam:
Well, the other solution could be to fix that in Qt, unless there's a valid reason for why they are like that. But then that would be for 5.9+, so i don't see a way other than having a blacklist with the issues to ignore.

krf:
ACK'd b giucam, merging.

PR #296, opened 2016-12-28
State Machines: Fix selection model syncing

krf:


krf:
@jpetersen Good point, I totally missed that. I've dropped	a55fd36 completely and adapted the "Fix selection model syncing" patch.

PR #301, opened 2017-01-09
Add CMake options to FeatureSummary

krf:
Simply provide a function wrapping both option() + add_feature_info():
  gammaray_option(...) -- drop-in replacement for option()

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,9 +133,9 @@ message(STATUS "Building GammaRay ${GAMMARAY_VERSION_STRING} in ${CMAKE_BUILD_TY
 #
 # Build options
 #
-option(
+gammaray_option(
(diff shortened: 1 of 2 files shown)


krf:
This will now print for instance:

```
-- The following features have been disabled:

 * Option GAMMARAY_ENFORCE_QT4_BUILD , Enforce a build with Qt4 (by default GammaRay looks out for Qt5 first)
 * Option GAMMARAY_PROBE_ONLY_BUILD , Build only an additional probe configuration for an already existing launcher.
 * Option GAMMARAY_ENABLE_GPL_ONLY_FEATURES , Enable features only available under GPL license.
 * Option GAMMARAY_INSTALL_QT_LAYOUT , Install into Qt directory layout.
 * Option GAMMARAY_MULTI_BUILD , Build multiple applicable probe configurations.
```

PR #304, opened 2017-01-20
Process tracker api proposal

pasnox:
Add api to possibility track the inspected process state

Being is process running ? is process debugged ? is debugged process paused or running ?

### DIFF
--- a/common/processtracker_macos.cpp
+++ b/common/processtracker_macos.cpp
@@ -1,0 +1,89 @@
+/*
+  processtracker_macos.cpp
+
+  This file is part of GammaRay, the Qt application inspection and
+  manipulation tool.
+
+  Copyright (C) 2016-2017 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Filipe Azevedo <filipe.azevedo@kdab.com>
+
+  Licensees holding valid commercial KDAB GammaRay licenses may use this file in
+  accordance with GammaRay Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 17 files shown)


common/processtracker_windows.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+    LocalBuffer(const LocalBuffer &other) = delete;
+    LocalBuffer &operator=(const LocalBuffer &other) = delete;
+
+    bool isValid() const {
+        return m_data != nullptr;
+    }
+
+    DWORD size() const {
+        return m_size;
```
giucam: If allocate() failed m_size will be different from 0 even if m_data is null, so maybe either set m_size to 0 there or check m_data here.

common/processtracker_macos.cpp
```diff
@@ -0,0 +1,86 @@
...
+        switch (info.kp_proc.p_stat) {
+            case SRUN:
+                pinfo.state = GammaRay::ProcessTracker::State::Running;
+                break;
+            case SSTOP:
+                pinfo.state = GammaRay::ProcessTracker::State::Suspended;
+                break;
+            default:
+                break;
+        }
```
giucam: Maybe print a warning if sysctl failed.

common/processtracker_macos.cpp
```diff
@@ -0,0 +1,86 @@
...
+    // Initialize mib, which tells sysctl the info we want, in this case
+    // we're looking for information about a specific process ID.
+    mib[0] = CTL_KERN;
+    mib[1] = KERN_PROC;
+    mib[2] = KERN_PROC_PID;
+    mib[3] = pid;
+
+    // Call sysctl.
+    size = sizeof(info);
+    junk = sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, NULL, 0);
```
pasnox: Done

common/processtracker_macos.cpp
pasnox: Done

common/processtracker_windows.cpp
pasnox: allocate() is private and can only result from resize() which first free() and set size to 0.
size it then updated only if ptr is not null.
Looks fine for me ?

giucam: Fair enough. Ok by me, then.

PR #311, opened 2017-02-02
Fix quickinspector (2commits)

chsterz:
Fixing bugs about an incomplete qquickitemmodel. 
(due to untracked scene intialization and the according windowchanges)

This makes the QuickItemModel more precise thus improving picking.

### DIFF
--- a/plugins/quickinspector/quickitemmodel.cpp
+++ b/plugins/quickinspector/quickitemmodel.cpp
@@ -281,6 +281,9 @@ void QuickItemModel::itemReparented()
 
     QQuickItem *sourceParent = m_childParentMap.value(item);
     Q_ASSERT(sourceParent);
+    if (sourceParent == item->parentItem())
+        return;
+
     const QModelIndex sourceParentIndex = indexForItem(sourceParent);
 
     QVector<QQuickItem *> &sourceSiblings = m_parentChildMap[sourceParent];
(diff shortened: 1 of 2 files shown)


plugins/quickinspector/quickitemmodel.cpp
```diff
@@ -156,7 +155,14 @@ void QuickItemModel::connectItem(QQuickItem *item)
...
 {
-    disconnect(item, nullptr, this, nullptr);
+    disconnect(item, SIGNAL(parentChanged(QQuickItem*)), this, SLOT(itemReparented()));
+    disconnect(item, SIGNAL(visibleChanged()), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(focusChanged(bool)), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(activeFocusChanged(bool)), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(widthChanged()), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(heightChanged()), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(xChanged()), this, SLOT(itemUpdated()));
+    disconnect(item, SIGNAL(yChanged()), this, SLOT(itemUpdated()));
```
pasnox: connectItem() is creating a new monitor each time it handle an item, but this does not delete it, resulting to possible multiple filters on the item.
Maybe we want to uninstall and delete the previous event filter here ?
Also the created monitor does have this as parent meaning the monitor will not be deleted until this is deleted, this let the monitor in memory when item is deleted for nothing.

vkrause: Good catch, one event filter per item is totally unnecessary here, We should get rid of that. But that's an unrelated improvement, so let's do that in a separate change.

PR #312, opened 2017-02-06
Allow showing only the current item in the Quick preview widget

giucam:
This adds a new button "Preview only the selected item" in the quickinspector ui, and when activated that makes the preview widget show only the item selected in the view, with its children, instead of the whole window. 
We probably need an icon for the button, because right now it's huge.


plugins/quickinspector/quickscenepreviewwidget.cpp
```diff
@@ -134,6 +136,17 @@ QuickScenePreviewWidget::QuickScenePreviewWidget(QuickInspectorInterface *inspec
 
     m_toolBar.toolbarWidget->addSeparator();
 
+    //TODO add icon
```
milianw: this needs to be solved, otherwise the GUI looks terrible

giucam: I agree, but making icons is not my strong point...

PR #314, opened 2017-02-09
Export the launcher dialog as a shared library

giucam:
This is to be used in the qtcreator plugin instead of the custom dialog.


vkrause:
When reviewing this I found it hard to reason about the dependencies in the launcher code, so I have shuffled the code around a bit and integrated the part of this change creating a shared library for the UI components (see wip/launcher-refactor branch).

Once that is merged we can rebase this change on top and should only be left with the API changes. I'm surprised though that there are no exports added in this change, which classes are actually needed in QtCreator in the end?

giucam:
I had forgotten to add the files with the new LauncherDialog class, which is the glue that was missing.

PR #317, opened 2017-02-10
Remove qtypetraits.h's contents altogether

jgueytat:
A recent version of modeltest.cpp is mandatory.

Read the following commit for full details:
https://codereview.qt-project.org/#/c/175770/13//ALL

"So that QFlags can use an (un)signed int matching the
underlying type as identified by the compiler and not by us.

Requires fixing a few warnings about sign conversion due to
QFlags misusages in qtbase that were either plain wrong, or
were relying on the enum being backed by an (un)signed int
when it wasn't."

### DIFF
--- a/3rdparty/qt/modeltest.cpp
+++ b/3rdparty/qt/modeltest.cpp
@@ -1,31 +1,26 @@
 /****************************************************************************
 **
-** Copyright (C) 2015 The Qt Company Ltd.
-** Contact: http://www.qt.io/licensing/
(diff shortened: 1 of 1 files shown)


krf:
Please fix the commit message, this should be rather like something along "Sync modeltest.cpp from upstream"

Feel free to link to the Gerrit CR in the commit message then.

jgueytat:
Commit message fixed

PR #320, opened 2017-02-15
Probe: Fix crash in signal callback handlers

krf:
Testing:
```
int main(int argc, char* argv[])
{
    QApplication app(argc, argv);

    new QTranslator;
    new QFileDialog;

    return 0;
}
```

Run:
  gammaray --inprocess --inject-only example

No longer crashes after this patch


krf:
Note: Adding the check `!Probe::instance()` is enough here, but I wanted to play safe and acquire the object lock in every method to get this code into a consistent shape.

I could revert the changes re. object lock if required.

krf:
<volker|office> kfunk: also, regarding the objectLock, IIRC I removed that explicitly there, as this is a performance hotpath and the lock is quite expensive

Found a different fix, will file a new pull request.

PR #322, opened 2017-02-15
Probe: Fix early-exit of debuggee

krf:
Start to add in-process integration tests

### DIFF
--- a/core/probe.cpp
+++ b/core/probe.cpp
@@ -282,9 +282,6 @@ QObject *Probe::window() const
 
 Probe *GammaRay::Probe::instance()
 {
-    if (!qApp)
-        return nullptr;
-
 #if (QT_VERSION < QT_VERSION_CHECK(5, 0, 0))
     return s_instance;
 #else
@@ -294,7 +291,7 @@ Probe *GammaRay::Probe::instance()
 
 bool Probe::isInitialized()
 {
-    return instance() && qApp;
+    return instance();
 }
 
 bool Probe::canShowWidgets()
@@ -309,6 +306,7 @@ bool Probe::canShowWidgets()
 
 void Probe::createProbe(bool findExisting)
 {
+    Q_ASSERT(qApp);
     Q_ASSERT(!isInitialized());
 
     // first create the probe and its children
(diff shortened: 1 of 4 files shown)


krf:
Supersedes #320

tests/CMakeLists.txt
```diff
@@ -208,6 +208,18 @@ endif()
 
 endif()
 
+### Integration test
+
+add_executable(integrationtest
+  integrationtest.cpp
+  ../probe/probecreator.cpp
```
milianw: why not link against gammaray_probe?

core/probe.cpp
```diff
@@ -396,6 +398,11 @@ void Probe::delayedInit()
         showInProcessUi();
 }
 
+void Probe::shutdown()
+{
+    delete this;
```
milianw: yikes ;-) usually, this is a very bad idea, but here there's really no way around this, is there?

core/probe.cpp
```diff
@@ -294,7 +291,7 @@ Probe *GammaRay::Probe::instance()
 
 bool Probe::isInitialized()
 {
-    return instance() && qApp;
+    return instance();
```
milianw: why is this change necessary?

core/probe.h
```diff
@@ -56,6 +56,14 @@ class BenchSuite;
 class Server;
 class ToolManager;
 
+/**
```
milianw: these comments could in principle be separate commits

core/probe.cpp
vkrause: The alternative I see for this is connect(qApp, &QCoreApp::aboutToQuit, [](){ delete Probe::instance(); }), but that's Qt5-only.

tests/CMakeLists.txt
krf: Copied from other tests in that file

core/probe.cpp
krf: `instance()` now implies that a `qApp` exists, thus this check is no longer necessary. I updated the documentation of the Probe to make sure it's clear there's a strong coupling between Probe & QCoreApplication lifetime.

core/probe.h
krf: There's no tight coupling between Probe lifetime & QCoreApplication lifetime before (because of the deleteLater), thus this comment does not make sense without the code changes.

core/probe.cpp
milianw: Reading http://stackoverflow.com/questions/550189/is-it-safe-to-delete-this I think it's fine to have it this way though

milianw:
+1 from my side, thanks a lot for the clarifications kevin

krf:
Pushed.

PR #329, opened 2017-03-01
Qml components traces feature

pasnox:


jpetersen:
Hi,

looks great in general, I noticed in QQ2 gallery example that something is wrong with the paint order when I have the menu overlay on the left open.

pasnox:
Indeed, I can reproduce, will fix !
Looks like the issue comes with popup or window overlay.

jpetersen:
Looks good

plugins/quickinspector/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if(Qt5Quick_FOUND AND HAVE_PRIVATE_Qt5Quick_HEADERS AND NOT ${Qt5Quick_VERSION}
   include_directories(SYSTEM ${Qt5Quick_PRIVATE_INCLUDE_DIRS})
 
   add_library(gammaray_quickinspector_shared STATIC ${gammaray_quickinspector_shared_srcs})
-  target_link_libraries(gammaray_quickinspector_shared gammaray_common Qt5::Gui Qt5::Quick)
+  target_link_libraries(gammaray_quickinspector_shared gammaray_common gammaray_core Qt5::Gui Qt5::Quick)
```
vkrause: This links probe code into the client, that is dangerous as that makes assumptions of running in target everywhere. Can we avoid this?

plugins/quickinspector/quickitemgeometry.cpp
```diff
@@ -112,6 +142,10 @@ void QuickItemGeometry::initFrom(QQuickItem *item)
         QQuickItemPrivate *parentPriv = QQuickItemPrivate::get(parent);
         parentTransform = parentPriv->itemToWindowTransform();
     }
+
+    traceColor = colorForItem(item);
+    traceTypeName = ObjectDataProvider::typeName(item);
+    traceName = ObjectDataProvider::name(item);
```
vkrause: This seems to be the only dependency on probe code in the shared bit, can probably be addressed by moving this to the caller instead.

plugins/quickinspector/CMakeLists.txt
pasnox: Good catch, will try

plugins/quickinspector/quickitemgeometry.cpp
pasnox: Yeah the ObjectDataProvider was the only required thing from the gammaray_core.
What do you mean by moving this to the caller ?

pasnox:
@vkrause Split the overlay/decoration drawing stuff now.

PR #331, opened 2017-03-10
Add generic Boot2Qt/Qt Automotive Suite Toolchain file

chsterz:
Use this toolchain file when a B2Qt./AS. environment setup file is present.
This is the case when obtaining image and sysroot from tqtc's online installer.

Reduces the required cmake call when crosscompiling to:
cmake -DCMAKE_TOOLCHAIN_FILE=<path/to/src>/cmake/Toolchain-Boot2Qt.cmake -DCMAKE_INSTALL_PREFIX= < install > <path/to/src>

### DIFF
--- a/cmake/Toolchain-Yocto.cmake
+++ b/cmake/Toolchain-Yocto.cmake
@@ -1,0 +1,43 @@
+# Basic cmake toolchain file for Qt for Yocto Environment
+# Assumptions: toolchain script is sourced
+#
+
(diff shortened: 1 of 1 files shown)


chsterz:
Andreas said this is a Yocto thing, so it might extend the applicability even to all crosscompilations using yocto toolchains 😄

vkrause:
+2, just change the copyright name before pushing :)

PR #333, opened 2017-03-13
Don't reset UI state when updating item or scenegraph tabs

chsterz:
Fixes: KDEND-140

### DIFF
--- a/plugins/quickinspector/quickinspectorwidget.cpp
+++ b/plugins/quickinspector/quickinspectorwidget.cpp
@@ -142,8 +142,6 @@ QuickInspectorWidget::QuickInspectorWidget(QWidget *parent)
     m_stateManager.setDefaultSizes(ui->mainSplitter, UISizeVector() << "50%" << "50%");
     m_stateManager.setDefaultSizes(ui->previewTreeSplitter, UISizeVector() << "50%" << "50%");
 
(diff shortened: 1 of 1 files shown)


chsterz:
Okay I put this on HOLD. PasNox said the error is further up the chain and these connects are actually needed.

vkrause:
This is still a very annoying issue, using the following workaround locally atm, which is less aggressive than the originally proposed solution:

diff --git a/plugins/quickinspector/quickinspectorwidget.cpp b/plugins/quickinspector/quickinspectorwidget.cpp
index 6eceda0..1512828 100644
--- a/plugins/quickinspector/quickinspectorwidget.cpp
+++ b/plugins/quickinspector/quickinspectorwidget.cpp
@@ -243,6 +243,8 @@ void GammaRay::QuickInspectorWidget::itemContextMenu(const QPoint &pos)
 
 void QuickInspectorWidget::resetState()
 {
+    disconnect(ui->itemPropertyWidget, SIGNAL(tabsUpdated()), this, SLOT(resetState()));
+    disconnect(ui->sgPropertyWidget, SIGNAL(tabsUpdated()), this, SLOT(resetState()));
     m_stateManager.reset();
 }

vkrause:
Fixed in cf389e5550c69551dd278bf6f31ede25fc9302b9.

PR #340, opened 2017-03-29
Frame transmission performance

chsterz:
This branch is the sum of a bit of work I invested in improving the frame transmission from weak embedded targets.

It features:
* LZ4-Compression by default, but still possible to disable on boards with a really weak CPU
* Categorized, enable-able Network traffic logging.
* A bugfix removing an artificial gap between frames
* A new method of reading frames (glReadPixels) 
  Which allows flipping GPU-read-images on the more powerful host instead of the weak target
* A supersecret (CTRL-Shift-RightClick) option in Quickinspector Contextmenu to output and visualize achieved  FPS rates

Many Thanks to Volker for that awesome debugging session :)

### DIFF
--- a/common/transferimage.h
+++ b/common/transferimage.h
@@ -44,6 +44,9 @@ public:
     const QImage &image() const;
     void setImage(const QImage &image);
 
+    QTransform transform() const;
+    void setTransform(const QTransform &transform);
+
     enum Format {
         QImageFormat,
         RawFormat
@@ -51,6 +54,7 @@ public:
 
 private:
     QImage m_image;
+    QTransform m_transform;
 };
 
 QDataStream &operator<<(QDataStream &stream, const GammaRay::TransferImage &image);
--- a/plugins/quickinspector/CMakeLists.txt
+++ b/plugins/quickinspector/CMakeLists.txt
@@ -15,8 +15,26 @@ if(Qt5Quick_FOUND AND HAVE_PRIVATE_Qt5Quick_HEADERS AND NOT ${Qt5Quick_VERSION} 
 
   include_directories(SYSTEM ${Qt5Quick_PRIVATE_INCLUDE_DIRS})
 
+  # Find if glReadPixels can be used with this configuration
+  # If yes, find based on which platform GL/GLESv2
+  if (TARGET Qt5::Gui_GLESv2)
+      set(HAVE_GLESv2 TRUE)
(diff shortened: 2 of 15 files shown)


CMakeLists.txt
```diff
@@ -285,7 +285,7 @@ endif()
 
 if(Qt5Core_FOUND)
   set_package_properties(Qt5Core PROPERTIES TYPE REQUIRED)
-  find_package(Qt5 NO_MODULE REQUIRED COMPONENTS Gui Network)
+  find_package(Qt5 NO_MODULE REQUIRED COMPONENTS Gui Network OpenGL)
```
vkrause: Considering we don't seem to use Qt5::OpenGL anywhere, do we really need it as a required dependency?

common/transferimage.h
```diff
@@ -51,6 +54,7 @@ class TransferImage
 
 private:
     QImage m_image;
+    bool m_flipped;
```
vkrause: Instead of the flip bit, could we make this a little bit more generic by using a QTransform instead? That would already pave the way for only sending a sub-rect of the frame for example?

plugins/quickinspector/CMakeLists.txt
```diff
@@ -15,8 +15,26 @@ if(Qt5Quick_FOUND AND HAVE_PRIVATE_Qt5Quick_HEADERS AND NOT ${Qt5Quick_VERSION}
...
+  find_package(OpenGL)
+  if (OpenGL_FOUND)
+      add_definitions(-DENABLE_GL_READPIXELS)
+  endif()
+
   add_library(gammaray_quickinspector_shared STATIC ${gammaray_quickinspector_shared_srcs})
-  target_link_libraries(gammaray_quickinspector_shared gammaray_common Qt5::Gui Qt5::Quick)
+  # on embedded devices link against GLESv2 for glReadpixels
+  if (HAVE_GLESv2)
+    target_link_libraries(gammaray_quickinspector_shared gammaray_common Qt5::Gui Qt5::Quick GLESv2)
```
vkrause: Shouldn't this be Qt5::Gui_GLESv2 instead here?

CMakeLists.txt
chsterz: Correct, I was mistaken by thinking headers like <QOpenGLContext> require QtOpenGL where in fact they come from Gui.

PR #341, opened 2017-03-31
Add new selector to ProbeInterface

pasnox:
Add new selector to ProbeInterface

This allow to select an object property.
Could be used by example to trigger a property selection in the Ivi Inspector.


pasnox:
The major ivi changes get cherry picked in master already.

PR #345, opened 2017-04-05
Master move traces

pasnox:
Move the interaction traces to visualize traces.
That's allowing to still use interaction (like measuring, redirect input...) while showing traces.


plugins/quickinspector/quickinspector.cpp
```diff
@@ -522,6 +525,15 @@ void QuickInspector::setCustomRenderMode(
 #else
     Q_UNUSED(customRenderMode);
 #endif
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
vkrause: This check is not needed here, QQ2 code cannot be built before Qt5 anyway.

pasnox: Ok, will remove

PR #346, opened 2017-04-07
Global UI rework, add themed icons set and client side models decorations

pasnox:
We now have 2 set of icons:

- Light
- Dark

Both comes with @2x and @3x dpi icons ensuring retina & up ready.
Also most server side decorations were moved to client side to avoid expensive network traffic.


vkrause:
Concept and approach look good to me.

I don't think we need to theme the class icons, they are not done by us but taken from Qt. They are not used for actions anyway, just as an additional visual indicator in tree views.

The use of QSettings in common looks problematic, we'd be touching the target configuration files there. Anything related to GammaRay's configuration should be contained inside the client and ui parts.

See also KDEND-151 for further considerations regarding probe-side icons, which would be another reason against adding icon lookup code in common.

pasnox:
Ho indeed, will fix all that.

ui/uiresources.h
```diff
@@ -0,0 +1,61 @@
...
+namespace GammaRay {
+namespace UIResources {
+    enum IconTheme {
+        Unknown,
+        Light,
+        Dark,
+        Default = Light
+    };
+
+    GAMMARAY_UI_EXPORT UIResources::IconTheme iconTheme();
```
vkrause: Does this need to be exported? Seems unused outside of the implementation of UIResources

ui/uiresources.h
```diff
@@ -0,0 +1,61 @@
...
+    GAMMARAY_UI_EXPORT void setIconTheme(UIResources::IconTheme theme);
+
+    GAMMARAY_UI_EXPORT QString themePath(UIResources::IconTheme theme);
```
vkrause: dito

ui/uiresources.h
```diff
@@ -0,0 +1,61 @@
...
+    GAMMARAY_UI_EXPORT QString themePath();
```
vkrause: seems unused?

ui/uiresources.h
```diff
@@ -0,0 +1,61 @@
...
+
+    GAMMARAY_UI_EXPORT QString themedPath(UIResources::IconTheme theme, const QString &extra);
+    GAMMARAY_UI_EXPORT QString themedPath(const QString &extra);
+
+    // Relative to CoreResources::themePath
+    GAMMARAY_UI_EXPORT QIcon themedIcon(UIResources::IconTheme theme, const QString &filePath);
```
vkrause: The above three methods also seem internally used only?

common/classesiconsindex_data.cpp
```diff
@@ -0,0 +1,139 @@
...
+namespace ClassesIconsIndex {
+static QHash<int, QString> s_numberToKey;
+static QHash<QString, int> s_keyToNumber;
+
+static void initializeClassesIconsIndex()
+{
+    Q_ASSERT(s_numberToKey.isEmpty());
+    Q_ASSERT(s_keyToNumber.isEmpty());
+    s_numberToKey[0] = QLatin1String(":/gammaray/icons/ui/classes/QQuickBorderImage/default.png");
+    s_keyToNumber[s_numberToKey[0]] = 0;
```
pasnox: I don't think it need to be part of the protocol.
It's just a simple int common to server and client which are always synced thanks to the generator tool and the protocol version matching.
It can be considered as any of the possible QAIM data role, just never hard code an icon id somewhere :)

common/utils.h
```diff
@@ -0,0 +1,68 @@
...
+ * @return a bool telling if the given color is light or not.
+ */
+GAMMARAY_COMMON_EXPORT bool isLightColor(const QColor &color);
+/**
+ * Returns true if the given color is dark, else false.
+ * @param color is the color to check.
+ *
+ * @return a bool telling if the given color is dark or not.
+ */
+GAMMARAY_COMMON_EXPORT bool isDarkColor(const QColor &color);
```
pasnox: It was what I had at first, the problem is that some model use it and not link to gammaray_ui (some model tests).

core/objectmodelbase.h
```diff
@@ -99,8 +99,8 @@ class ObjectModelBase : public Base
             return QVariant::fromValue(ObjectId(object));
         } else if (role == Qt::ToolTipRole) {
             return Util::tooltipForObject(object);
-        } else if (role == Qt::DecorationRole && index.column() == 0) {
-            return Util::iconForObject(object);
+        } else if (role == ObjectModel::DecorationIdRole && index.column() == 0) {
+            return Util::iconIdForObject(object);
```
pasnox: Ah indeed, I can change that.

ui/tools/objectinspector/objectinspectorwidget.cpp
```diff
@@ -52,12 +53,14 @@ ObjectInspectorWidget::ObjectInspectorWidget(QWidget *parent)
...
+    ClientDecorationIdentityProxyModel *clientModel = new ClientDecorationIdentityProxyModel(this);
+    clientModel->setSourceModel(model);
     ui->objectTreeView->header()->setObjectName("objectTreeViewHeader");
-    ui->objectTreeView->setModel(model);
+    ui->objectTreeView->setModel(clientModel);
     ui->objectTreeView->setContextMenuPolicy(Qt::CustomContextMenu);
     ui->objectTreeView->setDeferredResizeMode(0, QHeaderView::Stretch);
     ui->objectTreeView->setDeferredResizeMode(1, QHeaderView::Interactive);
-    new SearchLineController(ui->objectSearchLine, model);
+    new SearchLineController(ui->objectSearchLine, clientModel);
```
pasnox: I will have a look

ui/uiresources.h
pasnox: Indeed, it's currently not used outside.

ui/uiresources.h
pasnox: Changing the icon theme is used by QtCreator GammaRay integration at least, but not yet from GammaRay stand alone indeed.
Others are also only used internally and can be non exported.

ui/uiresources.h
pasnox: Yep

common/classesiconsindex_data.cpp
vkrause: one way might be to just send the complete map from the probe to the client, that's how we solve this for enums for example

common/utils.h
vkrause: Quite possible indeed, but I think we should rather link the tests to gammaray_ui in that case.

We have gotten requests from people wanting to use GammaRay on a QtCore-only device. That is currently not possible, but I'm at least trying to not move into the opposite direction with changes we make to the probe :)

Independent of that, it's a bit safer to not offer methods to the probe that might impact visual output on the client, somebody someday might use them ;-) We had quite some trouble in the past with dark color schemes on the client and light color schemes on the target device making the output unreadable.

pasnox: done

core/objectmodelbase.h
pasnox: done

ui/tools/objectinspector/objectinspectorwidget.cpp
pasnox: Good catch, it allow me to find and fix existing issues at some places with this search line controller!
Done

ui/uiresources.h
pasnox: done

ui/uiresources.h
pasnox: done

ui/uiresources.h
pasnox: done

ui/uiresources.h
pasnox: done

common/classesiconsindex_data.cpp
pasnox: done

pasnox:
All fixed, can give another round of review now.

PR #351, opened 2017-04-27
compensating for windows runtime linker deficiencies

photex:


photex:
Going to double check that this wouldn't also affect non-msvc builds. In which case I should check for a different variable than `MSVC`

photex:
Confirmed that at least for MingW builds, the MSVC var isn't set.

I'm on the fence about marking this as win32. It couldn't hurt, but if you're environment isn't completely initialized for Mingw already then you can't get anywhere near here because gcc won't work.

Meanwhile, if you generate a visual studio solution with cmake you'll require this to have a working build.

Thoughts?

common/CMakeLists.txt
```diff
@@ -7,11 +7,18 @@ if(NOT CMAKE_CROSSCOMPILING)
       target_link_libraries(classesiconsextractor LINK_PUBLIC ${QT_QTCORE_LIBRARIES} ${QT_QTXML_LIBRARIES})
     endif()
 
+    # Windows does not support rpath so in order to this command to succeed the Qt libraries must be in PATH
+    if (MSVC)
```
krf: As Filipe said: This is a problem on Windows in general -> `if (WIN32)`

common/CMakeLists.txt
```diff
@@ -7,11 +7,18 @@ if(NOT CMAKE_CROSSCOMPILING)
...
+    else()
+      set(PATH_PREFIX)
+    endif()
+
     add_custom_command(
         OUTPUT
             ${CMAKE_CURRENT_SOURCE_DIR}/classesiconsindex_data.cpp
         COMMAND
-            classesiconsextractor
+            ${PATH_PREFIX} classesiconsextractor
```
krf: Instead try this solution here: http://stackoverflow.com/questions/35029277/how-to-modify-environment-variables-passed-to-custom-cmake-target -- e.g.:

```
COMMAND ${CMAKE_COMMAND} -E env "PATH=C:/Some/Path;$ENV{PATH}" <real_command>
```

PR #354, opened 2017-05-03
MessageBuffer improvements

krf:
Introduce a shared object pool for the internal buffers of the Message
class. Message uses a pool of objects to read/write data into, this is
order to cut down the number of allocations done within Message.

Improvements:
- Cut down temporary allocations within QByteArray
- Cut down QDataStream/QBuffer creations to the size of the pool instead
  of the number of Messages created

Task-Id: KDEND-148

### DIFF
--- a/common/sharedpool.h
+++ b/common/sharedpool.h
@@ -1,0 +1,109 @@
+/*
+  This file is part of GammaRay, the Qt application inspection and
+  manipulation tool.
+
+  Copyright (C) 2017 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Kevin Funk <kevin.funk@kdab.com>
+
+  Licensees holding valid commercial KDAB GammaRay licenses may use this file in
+  accordance with GammaRay Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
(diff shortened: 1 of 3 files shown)


common/sharedpool.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    size_t size() const
+    {
+        return m_pool.size();
+    }
+
+private:
+    size_t m_capacity;
+    std::stack<std::unique_ptr<T> > m_pool;
```
akreuzkamp: wouldn't a std::stack using std::vector be faster than one using std::deque, given we don't need persistent pointers to the pointers? I really don't know, what's faster, thus consider it a question :)

krf: Good question. I think for this particular use case in doesn't matter much, since the capacity of the stack will be at most around 2-3 max, which fits into one of the fixed-size blocks of `std::dequeue`, thus there'll be minimal costs for memory allocation. When all the objects fit into one of the blocks you basically have the objects laid out in contiguous memory in a `std::deque`.

Anyway, looks like for an (almost) fixed-size stack, using `std::vector` as the backing container seems like the better choice. Right.

Thanks for the comment, made me look up the differences between those two container types :)

Some good discussion over here:
https://www.reddit.com/r/cpp/comments/2f7qzm/gotw_54_the_stdvector_vs_stddeque_debate_in/
And some implementation details about `std::deque`:
http://stackoverflow.com/questions/6292332/what-really-is-a-deque-in-stl/6292437#6292437

vkrause:
Looks good to me, but the CI didn't like this yet, right?

krf:
No CI failures afaics, will push then.

PR #355, opened 2017-05-04
Dynamically negotiate the more recent data stream version

pasnox:
Task-Id: KDEND-161


common/message.cpp
```diff
@@ -175,6 +175,45 @@ Message Message::readMessage(QIODevice *device)
...
 }
 
+quint8 Message::lowestSupportedDataVersion()
+{
+    return QDataStream::Qt_4_8;
+}
+
+quint8 Message::highestSupportedDataVersion()
+{
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
```
krf: I think we can just return `Qt_DefaultCompiledVersion` here.

From qdatastream.h:
```
#if QT_VERSION >= 0x050900
#error Add the datastream version for this Qt version and update Qt_DefaultCompiledVersion
#endif
        Qt_DefaultCompiledVersion = Qt_5_8
```

client/client.h
```diff
@@ -94,12 +94,13 @@ private slots:
...
 private:
     enum InitState {
-        VersionChecked = 1,
-        ObjectMapReceived = 2,
-        ServerInfoReceived = 4,
-        ConnectionEstablished = 8,
+        VersionChecked = 0x1,
+        ObjectMapReceived = 0x2,
+        ServerInfoReceived = 0x4,
+        ServerDataVersionNegociated = 0x8,
```
krf: Typo: `negociated`

pasnox: done

common/message.cpp
pasnox: Ah did not known about this one, thanks.

PR #362, opened 2017-05-18
Correctly handle IP when provided with --listen

chsterz:
* If IP cannot be used, send a message back to the Launcher
* Scan All devices for the user-specified IP, dont just take the first IP available on the first interface
* Avoids Printing bogus IPs on startup (sometimes 'Server is running on tcp:' was printed)

### DIFF
--- a/core/probe.cpp
+++ b/core/probe.cpp
@@ -416,8 +416,12 @@ void Probe::delayedInit()
     // The applicationName might be translated, so let's go with the application file base name
     m_server->setKey(QFileInfo(qApp->applicationFilePath()).completeBaseName());
     m_server->setPid(qApp->applicationPid());
-    m_server->listen();
-    ProbeSettings::sendServerAddress(m_server->externalAddress());
+    bool serverStarted = m_server->listen();
+    if (serverStarted) {
+        ProbeSettings::sendServerAddress(m_server->externalAddress());
+    } else {
+        ProbeSettings::sendServerLaunchError(m_server->errorString());
+    }
 
     if (ProbeSettings::value(QStringLiteral("InProcessUi"), false).toBool())
         showInProcessUi();
--- a/core/remote/tcpserverdevice.cpp
+++ b/core/remote/tcpserverdevice.cpp
@@ -61,6 +61,33 @@ bool TcpServerDevice::isListening() const
     return m_server->isListening();
 }
 
+QString TcpServerDevice::bestAvailableIP(const QHostAddress address) const
+{
+    QString firstHostFound;
+    foreach (const QNetworkInterface &inter, QNetworkInterface::allInterfaces()) {
+        if (!(inter.flags() & QNetworkInterface::IsUp)
+            || !(inter.flags() & QNetworkInterface::IsRunning)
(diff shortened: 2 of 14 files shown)


core/probe.cpp
```diff
@@ -416,8 +416,13 @@ void Probe::delayedInit()
     // The applicationName might be translated, so let's go with the application file base name
     m_server->setKey(QFileInfo(qApp->applicationFilePath()).completeBaseName());
     m_server->setPid(qApp->applicationPid());
-    m_server->listen();
-    ProbeSettings::sendServerAddress(m_server->externalAddress());
+    bool serverStarted = m_server->listen();
+    if (serverStarted)
+    {
```
jpetersen: no newline

core/remote/tcpserverdevice.cpp
```diff
@@ -61,6 +61,33 @@ bool TcpServerDevice::isListening() const
     return m_server->isListening();
 }
 
+QString TcpServerDevice::bestAvailableIP(const QHostAddress address) const
+{
+    QString result;
```
jpetersen: rename to firstHostFound or so

core/remote/tcpserverdevice.cpp
```diff
@@ -61,6 +61,33 @@ bool TcpServerDevice::isListening() const
...
+
+        foreach (const QNetworkAddressEntry &addrEntry, inter.addressEntries()) {
+            const QHostAddress addr = addrEntry.ip();
+
+            // Return the ip according to the listening server protocol.
+            if (addr.protocol() != m_server->serverAddress().protocol()
+                || !addr.scopeId().isEmpty())
+                continue;
+
+            if (result.isEmpty())
```
jpetersen: switch both if statements (do not need to store in result if we return anyways)

core/probe.cpp
```diff
@@ -416,8 +416,13 @@ void Probe::delayedInit()
...
+        ProbeSettings::sendServerAddress(m_server->externalAddress());
+    } else {
+        ProbeSettings::sendServerLaunchError("The address is not available");
```
vkrause: Maybe send the QTcpServer error string here, instead the hardcoded error message?

chsterz:
[Note to self] What I should do also: 
 - Handle the standard case (no --listen provided) such that if the launcher is started, localhost is used and use 0.0.0.0 else. 
 - If 0.0.0.0 is used, list all of the interfaces with all their IPs avail as a convenience to the user.

launcher/core/launcher.cpp
```diff
@@ -329,6 +330,13 @@ void Launcher::readyRead()
             msg >> d->serverAddress;
             break;
         }
+        case Protocol::ServerLaunchError:
+        {
+            QString reason;
+            msg >> reason;
+            QMessageBox::critical(nullptr, "GammaRay server could not start", "The server could not start: " + reason);
```
vkrause: This looks problematic in terms of dependencies, as the core part of the launcher code can also used without widgets even being available, e.g. think about the CLI launcher on an embedded target.

chsterz: had that thought too, but must have forgotten about it when it worked so nicely ;)
I'll remove

PR #364, opened 2017-05-23
Perf: Improve iconIdForObject()...

krf:
Reduce allocations caused by QByteArray::fromRawData(...)

Task-Id: KDEND-171

### DIFF
--- a/core/util.cpp
+++ b/core/util.cpp
@@ -141,6 +141,8 @@ struct IconCacheEntry
         : defaultIcon(-1)
     { }
 
+    /// note: this member's data is being referenced by the database's key
+    QByteArray className;
     int defaultIcon;
 
     // pair of property name and expected string value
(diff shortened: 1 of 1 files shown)


core/util.cpp
```diff
@@ -167,7 +169,9 @@ static IconDatabase readIconData()
         const QDir dir(fileInfo.path());
         const QString className = dir.dirName();
         const QString iconName = fileInfo.fileName();
-        IconCacheEntry &perClassData = data[className.toLatin1()];
+        const QByteArray classNameAsByteArray = className.toLatin1();
+        IconCacheEntry &perClassData = data[QLatin1String(classNameAsByteArray)];
```
marc-kdab: You might want to pass `data()` + `size()` here, as the `QLatin1String(QByteArray)` ctor performs a (useless) `strlen()`.

core/util.cpp
```diff
@@ -167,7 +169,9 @@ static IconDatabase readIconData()
...
+        perClassData.className = classNameAsByteArray;
```
marc-kdab: std::move ?

core/util.cpp
krf: Fixed.

core/util.cpp
krf: Fixed.

PR #369, opened 2017-06-06
TimerTop thread safety

pasnox:
Timer inspector enhancements

### DIFF
--- a/core/probe.cpp
+++ b/core/probe.cpp
@@ -128,6 +128,7 @@ static void signal_end_callback(QObject *caller, int method_index)
     QMutexLocker locker(Probe::objectLock());
     if (!Probe::instance()->isValidObject(caller)) // implies filterObject()
         return; // deleted in the slot
+    locker.unlock();
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
     method_index = Util::signalIndexToMethodIndex(caller->metaObject(), method_index);
@@ -157,6 +158,7 @@ static void slot_end_callback(QObject *caller, int method_index)
     QMutexLocker locker(Probe::objectLock());
     if (!Probe::instance()->isValidObject(caller)) // implies filterObject()
         return; // deleted in the slot
+    locker.unlock();
 
     Probe::executeSignalCallback([=](const SignalSpyCallbackSet &callbacks) {
             if (callbacks.slotEndCallback)
--- a/plugins/timertop/timerinfo.cpp
+++ b/plugins/timertop/timerinfo.cpp
@@ -30,194 +30,229 @@
 #include "timermodel.h"
 
 #include <core/util.h>
+#include <core/probe.h>
 
 #include <QObject>
+#include <QTimer>
+#include <QThread>
+#include <QAbstractEventDispatcher>
+#include <QMutexLocker>
 
 using namespace GammaRay;
 
-static const int maxTimeoutEvents = 1000;
-static const int maxTimeSpan = 10000;
-
-TimerInfo::TimerInfo(QObject *timer)
-    : m_type(QQmlTimerType)
-    , m_totalWakeups(0)
-    , m_timer(timer)
-    , m_timerId(-1)
-    , m_lastReceiver(nullptr)
+namespace GammaRay {
+uint qHash(const TimerId &id)
 {
-    if (QTimer *t = qobject_cast<QTimer *>(timer)) {
+    switch (id.m_type) {
+    case TimerId::InvalidType:
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+        Q_UNREACHABLE();
+#else
+        Q_ASSERT(false);
+#endif
+        break;
+
+    case TimerId::QQmlTimerType:
+    case TimerId::QTimerType:
+        return ::qHash(id.m_timerAddress);
+
+    case TimerId::QObjectType:
+        return ::qHash(id.m_timerId);
+    }
+
+    return 0;
+}
+}
+
+TimerId::TimerId()
+    : m_type(InvalidType)
+    , m_timerAddress(0)
+    , m_timerId(-1)
+{
+}
+
+TimerId::TimerId(QObject *timer)
+    : m_type(QQmlTimerType)
+    , m_timerAddress(quintptr(timer))
+    , m_timerId(-1)
+{
+    Q_ASSERT(timer);
+
+    if (qobject_cast<QTimer *>(timer)) {
         m_type = QTimerType;
-        m_timerId = t->timerId();
     }
 }
 
-TimerInfo::TimerInfo(int timerId)
+TimerId::TimerId(int timerId)
     : m_type(QObjectType)
-    , m_totalWakeups(0)
+    , m_timerAddress(0)
     , m_timerId(timerId)
 {
+    Q_ASSERT(m_timerId != -1);
 }
 
-TimerInfo::Type TimerInfo::type() const
+TimerId::Type TimerId::type() const
 {
     return m_type;
 }
 
-void TimerInfo::addEvent(const TimeoutEvent &timeoutEvent)
+quintptr TimerId::address() const
 {
-    m_timeoutEvents.append(timeoutEvent);
-    removeOldEvents();
-    m_totalWakeups++;
+    return m_timerAddress;
 }
 
-int TimerInfo::numEvents() const
-{
-    return m_timeoutEvents.size();
-}
-
-QObject *TimerInfo::timerObject() const
-{
-    return m_timer;
-}
-
-QTimer *TimerInfo::timer() const
-{
-    if (m_type != QTimerType)
-        return nullptr;
-    return qobject_cast<QTimer *>(m_timer);
-}
-
-int TimerInfo::timerId() const
+int TimerId::timerId() const
 {
     return m_timerId;
 }
 
-FunctionCallTimer *TimerInfo::functionCallTimer()
+bool TimerId::operator==(const TimerId &other) const
 {
-    return &m_functionCallTimer;
-}
+    if (m_type != other.m_type)
+        return false;
 
-QString TimerInfo::wakeupsPerSec() const
-{
-    int totalWakeups = 0;
-    int start = 0;
-    int end = m_timeoutEvents.size() - 1;
-    for (int i = end; i >= 0; i--) {
-        const TimeoutEvent &event = m_timeoutEvents.at(i);
-        if (event.timeStamp.msecsTo(QTime::currentTime()) > maxTimeSpan) {
-            start = i;
-            break;
-        }
-        totalWakeups++;
-    }
-
-    if (totalWakeups > 0 && end > start) {
-        const QTime startTime = m_timeoutEvents[start].timeStamp;
-        const QTime endTime = m_timeoutEvents[end].timeStamp;
-        const int timeSpan = startTime.msecsTo(endTime);
-        const float wakeupsPerSec = totalWakeups / (float)timeSpan * 1000.0f;
-        return QString::number(wakeupsPerSec, 'f', 1);
-    }
-    return QStringLiteral("0");
-}
-
-QString TimerInfo::timePerWakeup() const
-{
-    if (m_type == QObjectType)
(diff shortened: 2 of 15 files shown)


plugins/timertop/timermodel.cpp
```diff
@@ -27,190 +27,387 @@
...
+void TimerModel::triggerPushChanges()
 {
-    if (!s_timerModel)
-        s_timerModel = new TimerModel;
-    Q_ASSERT(s_timerModel);
-    return s_timerModel;
+    if (!TimerModel::isInitialized())
+        return;
+
+    QMutexLocker locker(Probe::objectLock());
```
vkrause: Why are we holding the object lock here, we are not checking QObject validity anywhere from what I can see?

plugins/timertop/timermodel.cpp
```diff
@@ -27,190 +27,387 @@
...
+    infoHash.squeeze();
 
-    // Create a new free timer, and emit the correct update signals
-    TimerInfoPtr timerInfo(new TimerInfo(timerId));
-    beginInsertRows(QModelIndex(), rowCount(), rowCount());
-    m_freeTimers.append(timerInfo);
-    endInsertRows();
-    return timerInfo;
+    s_timerModel->m_applyChangesMethod.invoke(s_timerModel, Qt::QueuedConnection,
+                                              Q_ARG(GammaRay::TimerIdInfoHash, infoHash));
```
vkrause: Why the extra queued invocation here? pushChanges() is already in the thread the model belongs too, and called from the event loop, no?

plugins/timertop/timermodel.cpp
```diff
@@ -27,190 +27,387 @@
...
+        const QTimer *const timer = qobject_cast<QTimer *>(receiver);
+
+        // If there is a QTimer associated with this timer ID, don't handle it here, it will be handled
+        // by the signal hooks preSignalActivate/postSignalActivate.
+        if (timer && timer->timerId() == timerEvent->timerId()) {
+            return false;
+        }
+
+        {
+            QMutexLocker locker(Probe::objectLock());
```
vkrause: I guess the lock is needed to protect s_gatheredTimersData? In that case we probably should use our own lock here rather than abuse the probe's object lock?

plugins/timertop/timermodel.cpp
```diff
@@ -317,37 +540,175 @@ QMap<int, QVariant> TimerModel::itemData(const QModelIndex &index) const
...
+        ++inserted;
+    }
+
+    // Inform model about new rows
+    if (inserted > 0) {
+        const int first = m_sourceModel->rowCount() + m_freeTimersInfo.count() - inserted;
+        const int last = m_sourceModel->rowCount() + m_freeTimersInfo.count() - 1;
+
+        beginInsertRows(QModelIndex(), first, last);
+        // We already inserted at end
```
vkrause: Does this still pass modeltest? This looks like it breaks some rowCount invariants in there.

plugins/timertop/timerinfo.h
```diff
@@ -28,67 +28,85 @@
...
+    virtual ~TimerIdInfo() { }
 
-    void removeOldEvents();
+    virtual void update(const TimerId &id, QObject *receiver = nullptr);
+
+    int timerId;
+    int totalWakeups;
+
+    QString objectName;
+    QString state;
```
vkrause: Does this (and the following 3) really need to be strings? Looking at moving tr() calls to the client, it would be better to have this as numeric values here, and only do the string conversion in the model on demand I think.

tests/timertoptest.cpp
```diff
@@ -74,84 +74,107 @@ class TimerTopTest : public QObject
     }
 
 private slots:
-    void testTimerCreateDestroy()
```
pasnox: Fixed

plugins/timertop/timermodel.h
```diff
@@ -68,63 +70,54 @@ class TimerModel : public QAbstractTableModel
...
-    // pending dataChanged() signals
-    QSet<int> m_pendingChangedTimerObjects;
-    QSet<int> m_pendingChangedFreeTimers;
-    QTimer *m_pendingChanedRowsTimer;
     // the method index of the timeout() signal of a QTimer
     const int m_timeoutIndex;
     int m_qmlTimerTriggeredIndex;
+
+    const QMetaMethod m_pushTimerStart;
+    QSharedPointer<QTimer> m_pushTimer;
```
pasnox: Fixed

plugins/timertop/timermodel.cpp
pasnox: Fixed

plugins/timertop/timermodel.cpp
pasnox: Fixed

plugins/timertop/timermodel.cpp
pasnox: Fixed

plugins/timertop/timermodel.cpp
pasnox: Fixed

core/probe.cpp
```diff
@@ -543,6 +543,14 @@ QMutex *Probe::objectLock()
     return s_lock();
 }
 
+QSharedPointer<QMutexLocker> Probe::objectLocker(QObject *object)
```
pasnox: Fixed

core/probe.cpp
```diff
@@ -543,6 +543,14 @@ QMutex *Probe::objectLock()
...
+{
+    Q_ASSERT(object);
+    if (object->thread() == QThread::currentThread())
+        return QSharedPointer<QMutexLocker>();
```
pasnox: Fixed

plugins/timertop/timerinfo.cpp
```diff
@@ -30,194 +30,253 @@
...
+void TimerIdInfo::update(const TimerId &id, QObject *receiver)
 {
-    m_lastReceiver = receiver;
-}
+    Q_ASSERT(id.isValid());
+    QObject *object = receiver ? receiver : id.timerObject();
+    // We can't lock TimerModel::m_mutex and Probe::objectLock() on the same thread at the same time
+    QSharedPointer<QMutexLocker> locker;
+    if (object->thread() != QThread::currentThread())
+        locker = QSharedPointer<QMutexLocker>(new QMutexLocker(Probe::objectLock()));
```
marc-kdab: Are you sure you don't want to use the movable `std::unique_lock` instead of this abomination which allocates memory twice?

vkrause: Even implementation details aside, locking a mutex only in some threads would seem problematic from a correctness point-of-view. The comment just reads like "otherwise it deadlocks" to me, if there are pre-conditions that make this code correct (like objectLock already being held by the caller in the main thread), this should be mentioned.

plugins/timertop/timerinfo.h
pasnox: Fixed

plugins/timertop/timerinfo.cpp
pasnox: Fixed

plugins/timertop/timertopwidget.cpp
```diff
@@ -40,6 +40,74 @@
...
+    }
+
+    static QString stateToString(uint state)
+    {
+        const int timerState = (state & 0xffff);
+        const int timerInterval = (state >> 16);
+
+        switch (timerState) {
+        case 0: // None
+            return QCoreApplication::translate("TimerModel", "None");
```
vkrause: This should be tr(), avoids manually passing the context (which would need to include the namespace prefix btw), but more importantly, only tr() is extracted for translation, and I think you removed the tr() calls in the model previously providing the strings.

plugins/timertop/timertopwidget.cpp
```diff
@@ -40,6 +40,74 @@
...
+        case 3: // Repeat
+            return QCoreApplication::translate("TimerModel", "Repeating (%1 ms)").arg(timerInterval);
+        }
+
+        return QString();
+    }
+
+    static QString wakeupsPerSecToString(qreal value)
+    {
+        return qFuzzyIsNull(value) ? QStringLiteral("0") : QString::number(value, 'f', 1);
```
vkrause: tr() for the simple constants here and below too

plugins/timertop/timerinfo.h
```diff
@@ -28,67 +28,89 @@
...
+    { }
+
+    virtual ~TimerIdInfo() { }
+
+    virtual void update(const TimerId &id, QObject *receiver = nullptr);
+    void lockAndUpdate(const TimerId &id, QObject *receiver = nullptr);
+
+    int timerId;
+    uint totalWakeups;
+    quint64 lastReceiverAddress; // The QTimer/QQmlTimer or last known receiver address
```
vkrause: If this will contain a pointer value (looks like it), prefer quintptr.

plugins/timertop/timermodel.cpp
```diff
@@ -27,190 +27,358 @@
...
 void TimerModel::preSignalActivate(QObject *caller, int methodIndex)
 {
-    if (!(methodIndex == m_timeoutIndex && qobject_cast<QTimer *>(caller))
-        && !(methodIndex == m_qmlTimerTriggeredIndex && caller->inherits("QQmlTimer")))
-        return;
+    // We are in the thread of the caller emmiting the signal
+    // The probe did NOT locked the objectLock at this point.
+    Q_ASSERT(TimerModel::isInitialized());
+
+    QMutexLocker locker(&s_mutex);
```
vkrause: Can we move this down to after the check if this is a relevant signal for us? Up to there this is performance-sensitive hot-path code (called for every single signal emission!), so we don't want to serialize this, nor add the cost of the mutex operation here if we can help it.

plugins/timertop/timerinfo.h
```diff
@@ -28,67 +28,89 @@
...
-typedef QSharedPointer<TimerInfo> TimerInfoPtr;
+typedef QHash<TimerId, TimerIdInfo> TimerIdInfoHash;
+
+uint qHash(const TimerId &id);
 }
 
-Q_DECLARE_METATYPE(GammaRay::TimerInfoPtr)
+Q_DECLARE_METATYPE(GammaRay::TimerId)
+Q_DECLARE_METATYPE(GammaRay::TimerIdInfo)
+Q_DECLARE_METATYPE(GammaRay::TimerIdInfoHash)
```
vkrause: Do we need all of those? Meta type declarations for GammaRay types "leak" into the target process, so we should keep them to the bare minimum.

plugins/timertop/timertopwidget.cpp
pasnox: As far as I know QApplication::translate() calls are extracted by lupdate too (See qthelp://org.qt-project.qtcore.551/qtdoc/i18n-source-translation.html).
You are right though - I need to fix the namespace thing.
I can not use TimerModel::tr() any longer as it depend QMetaObject and is not compiled/linked in the client.

plugins/timertop/timerinfo.h
pasnox: Hm xD it's what I used at first and moved to quint64 because it's what use ObjectId.

plugins/timertop/timerinfo.h
pasnox: Ho, indeed I did not though about that.. will check.

plugins/timertop/timertopwidget.cpp
pasnox: Fixed

plugins/timertop/timertopwidget.cpp
pasnox: Fixed

plugins/timertop/timerinfo.h
pasnox: Fixed

plugins/timertop/timerinfo.h
pasnox: Fixed

plugins/timertop/timermodel.cpp
pasnox: Fixed

PR #370, opened 2017-06-07
Perfomance fixes for quickinspector

krf:


vkrause:
Looks good to me. Did you measure this to make sure the costs of the new code are actually lower? They should be AFAICT, but you never know ;-)

milianw:
the sender optimization looks good to me - that call is linear in the number of connections otherwise and should now be O(1). Still, hard numbers would be good to see.

The other one is I think a bit of a misnomer - a hover event per se isn't expensive, it just happens far too often which makes it expensive to send over the wire, right? So maybe make that clear with a comment at least?

krf:
@milianw Added a benchmark + put benchmark diff in the commit msg. Also fixed the misnomer.

Now: Trying to get my stuff compiled on MSVC2010 again...

krf:
Fixed compilation under MSVC2010, code got slightly uglier.

plugins/quickinspector/quickitemmodel.cpp
```diff
@@ -135,34 +135,42 @@ void QuickItemModel::populateFromItem(QQuickItem *item)
     QVector<QQuickItem *> &children = m_parentChildMap[item->parentItem()];
     std::sort(children.begin(), children.end());
 
-    // Make sure every items are known to the objects model as this is not always
-    // the case when attaching to running process (OSX)
-    Probe::instance()->discoverObject(item);
+    if (Probe::instance()) {
+        // Make sure every items are known to the objects model as this is not always
```
pasnox: You can remove this comments.
It clear that when attaching, one need discoverObject on any platform. is not it ?

krf: Could be done in another patch, yep. I'm just moving those lines here.

krf:
Rebased. Please review.

PR #371, opened 2017-06-07
Show transmission rate in status bar in mainwindow

krf:


krf:
gammaray-transmissionrate-in-mainwindow

krf:
Comments? Much more visible than in the debug output IMO. Could be enabled conditionally by an environment variable.

vkrause:
I agree. Only useful for GammaRay developers though, not for most of the users, so hiding this by default is probably a good idea indeed.

krf:
Needs the following env var now: `GAMMARAY_DEVELOPERMODE=1`

common/endpoint.h
```diff
@@ -255,7 +257,8 @@ private slots:
 
     QPointer<QIODevice> m_socket;
     Protocol::ObjectAddress m_myAddress;
-    int m_bytesTransferred;
+    int m_bytesRead;
+    int m_bytesWritten;
```
vkrause: uint64 while we are at it please, we already managed to overflow those counters

ui/mainwindow.cpp
```diff
@@ -313,6 +313,14 @@ MainWindow::MainWindow(QWidget *parent)
     connect(UiIntegration::instance(), SIGNAL(navigateToCode(QUrl,int,int)), this,
             SLOT(navigateToCode(QUrl,int,int)));
 
+    const bool developerModeEnabled = !QString::fromLocal8Bit(qgetenv("GAMMARAY_DEVELOPERMODE")).isEmpty();
```
vkrause: Nice, @chsterz we probably want to use that for the remote view FPS option too. Same for the message statistics view.

PR #373, opened 2017-06-12
Disable old QIviProperty based test on newer versions of QtIVI

mkrus:
needs to be replaced with a newer test


vkrause:
+2, but please push to the 2.8 branch so we have tests passing there too. Thanks!

PR #377, opened 2017-06-23
Signal inspector optimizations and improvements

pasnox:


plugins/signalmonitor/signalhistorymodel.cpp
```diff
@@ -156,14 +172,18 @@ QVariant SignalHistoryModel::data(const QModelIndex &index, int role) const
...
 {
-    if (role == Qt::DisplayRole && orientation == Qt::Horizontal) {
-        switch (section) {
-        case ObjectColumn:
-            return tr("Object");
-        case TypeColumn:
-            return tr("Type");
-        case EventColumn:
-            return tr("Events");
+    if (orientation == Qt::Horizontal) {
```
milianw: the two conditionals can be merged like before, resulting in a decreased indent-depth

plugins/signalmonitor/signalhistorymodel.cpp
```diff
@@ -182,6 +202,31 @@ QMap< int, QVariant > SignalHistoryModel::itemData(const QModelIndex &index) con
     return d;
 }
 
+bool SignalHistoryModel::setData(const QModelIndex &index, const QVariant &value, int role)
+{
+    switch (static_cast<ColumnId>(index.column())) {
```
milianw: considering only one column is editable, why not use an if here instead of a switch?

plugins/signalmonitor/signalmonitor.cpp
```diff
@@ -40,14 +40,32 @@
...
+
+class SignalMonitorProxyModel : public ServerFilterProxyModel
+{
+public:
+    using ServerFilterProxyModel::ServerFilterProxyModel;
+
+    void sort(int column, Qt::SortOrder order = Qt::AscendingOrder) override
+    {
+        if (static_cast<SignalHistoryModel::ColumnId>(column) == SignalHistoryModel::MonitoredColumn) {
+            setSortRole(Qt::CheckStateRole);
```
milianw: I guess you tested this, but I wonder why this isn't triggering a recursion ;-) the proxy is dynamic, and changing the sort role should resort? but apparently not - good to know

plugins/signalmonitor/signalhistoryview.cpp
```diff
@@ -58,11 +58,11 @@ SignalHistoryView::SignalHistoryView(QWidget *parent)
     connect(m_eventDelegate, SIGNAL(visibleIntervalChanged(qint64)),
             this, SIGNAL(delegateVisibleIntervalChanged(qint64)));
     connect(m_eventDelegate, SIGNAL(visibleOffsetChanged(qint64)),
-            this, SLOT(eventDelegateChanged()));
+            this, SLOT(eventDelegateChanged()), Qt::QueuedConnection);
```
milianw: Qt does _not_ group similar events, so this whole change set is highly dubious to me. If you want to group stuff, you'll have to implement the grouping yourself.

plugins/signalmonitor/signalhistorydelegate.cpp
```diff
@@ -67,7 +67,7 @@ void SignalHistoryDelegate::paint(QPainter *painter, const QStyleOptionViewItem
     const qint64 endTime = startTime + interval;
 
     const QAbstractItemModel * const model = index.model();
-    const QVector<qint64> &events
+    const QVector<qint64> events
```
milianw: this change is not required

plugins/signalmonitor/signalhistorydelegate.cpp
```diff
@@ -95,10 +95,15 @@ void SignalHistoryDelegate::paint(QPainter *painter, const QStyleOptionViewItem
 
     foreach (qint64 ev, events) {
         const qint64 ts = SignalHistoryModel::timestamp(ev);
-        if (ts >= startTime && ts < endTime) {
-            const int x = x0 + dx * (ts - startTime) / interval;
-            lines << QLine(x, y0 + 1, x, y0 + dy - 2);
-        }
+
```
milianw: how is this optimizing the rendering? the code is exactly the same as before, you just split the conditoinal in two - this cannot possibly influence the performance?!

plugins/signalmonitor/signalhistorydelegate.cpp
```diff
@@ -49,7 +49,7 @@ SignalHistoryDelegate::SignalHistoryDelegate(QObject *parent)
     , m_totalInterval(0)
 {
     connect(m_updateTimer, SIGNAL(timeout()), this, SLOT(onUpdateTimeout()));
-    m_updateTimer->start(1000 / 25);
+    m_updateTimer->start(1000 / 15);
```
milianw: you are not "optimizing the frame per seconds", you are "reducing the frames per second".

milianw: also, the next commit obsoletes this one, you can drop it altogether

plugins/signalmonitor/signalmonitor.cpp
pasnox: The sort implementation is done in the dptr, setSortRole() only call d->sort(), not QAIM::sort().

plugins/signalmonitor/signalhistoryview.cpp
pasnox: Hm indeed I though identical queued event would ALL be compressed, but from reading the code at https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qcoreapplication.cpp.html#1442 it seems very few are in reality.
Thanks for pointing this out!

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: This all rely on the QPainter implementation.
It is far faster drawing multiple lines with QPainter::drawLines() than multiple QPainter::drawLine().

pasnox:
Indeed my metrics is not really *accurate*, just checked the cpu usage, and when you pass from 120% usage to 60% i guess you can trust it a bit ;)
I will try to add some benchmarks, though I did not yet played with that thing.

plugins/signalmonitor/signalhistorydelegate.cpp
milianw: Yes, but even before you used QPainter::drawLines. Just read the diff! You are doing one and the same thing on both sides, just formatted differently. This cannot possibly influence the performance.

milianw:
I'm not debating that the patches have a good influence on the performance overall. I'm just saying that some patches are dubious. And you need to better explain what metric you are using to justify the patches. From what I now gather, you looked at the CPU time in the Mac-equivalent of "top"? That is super inaccurate on its own. But more importantly, this means that you have had to "setup" GammaRay and whichever test application you used manually everytime between the measurements, making the measurement even more inaccurate.

Please, at least use a CPU profiler like Instruments to do the measurement. Then you can hopefully make some more accurate statements on individual functions.

plugins/signalmonitor/signalhistoryview.cpp
pasnox: Dropped

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: Done

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: Ho you highlighted the timestamps checks... sorry. I tough it was about the drawLine / drawLines.
This change is minor and very specific while you scroll and the history length is wide.
Before it would continue reading past the endTime events, while now it break when it reach the first past endTime event, so no, it does not exactly do the same thing as before.

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: Dropped.

ui/deferredtreeview.cpp
```diff
@@ -56,6 +56,17 @@ void setSectionResizeMode(QHeaderView *header, int logicalIndex, QHeaderView::Re
...
+QStyle::StateFlag checkStateStyleFlag(Qt::CheckState state) {
+    switch (state) {
+    case Qt::Unchecked:
+        return QStyle::State_Off;
+    case Qt::PartiallyChecked:
+        return QStyle::State_NoChange;
+    case Qt::Checked:
+        return QStyle::State_On;
+    }
+}
```
chsterz: GCC leaves me with a 
`/home/christoph/KDAB/QT/GammaRay/ui/deferredtreeview.cpp:69:1: error: control reaches end of non-void function [-Werror=return-type]` here. Maybe add a default?

pasnox: Fixed

plugins/signalmonitor/signalhistorydelegate.cpp
```diff
@@ -47,10 +47,10 @@ SignalHistoryDelegate::SignalHistoryDelegate(QObject *parent)
     , m_visibleOffset(0)
     , m_visibleInterval(15000)
     , m_totalInterval(0)
+    , m_fps(0)
 {
     connect(m_updateTimer, SIGNAL(timeout()), this, SLOT(onUpdateTimeout()));
-    m_updateTimer->start(1000 / 25);
-    onUpdateTimeout();
+    setFPS(15);
```
vkrause: Isn't this addressing the problem on the wrong side? Shouldn't we rather rate-limit the dataChanged signals in the signal history model already, like we do in the meta object model? Maybe that rate limiter should even move to RemoteModelServer generically? Performance on the client isn't really out problem after all, the problem is poor performance on the target, excessive client operations are usually merely a symptom of that.

plugins/signalmonitor/signalhistorymodel.h
```diff
@@ -64,6 +65,7 @@ class SignalHistoryModel : public QAbstractTableModel
 
 public:
     enum ColumnId {
+        MonitoredColumn = 0,
```
vkrause: The extra column needs a lots of space for just the checkbox, can we merge this into the object name column maybe?

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: That would indeed help more also.
Still if you look closer in the code you will see that now no network traffic should be emitted for the events by default (You have to tick an object row so it become monitored and start transmitting signals data). The amont of data sent by default should not be more than the object one itself.
What is still an issue is the high frequency clock time of the signal monitor interface this is it struggling the network traffic.

plugins/signalmonitor/signalhistorymodel.h
pasnox: Its width is roughly "X M. ^", so yeah something like 3 times of the width of the check indicator itself, not so much in my opinion, still it depend which screen size you use :)
This column add some benefit to the view, ie, you can sort by checked rows and still be able to sort by name.
That can be useful to uncheck rows that are not contiguous, it can become very boring to scroll all the view to find the missing checked row (ok, you can click the header check indicator to check all then unselect all... at the cost of some network traffic).
If it really broke your daily confort usage within this inspector, then of course we can merge it.

core/remote/remotemodelserver.cpp
```diff
@@ -438,6 +478,7 @@ void RemoteModelServer::layoutChanged()
 void RemoteModelServer::layoutChanged(const QList<QPersistentModelIndex> &parents,
                                       QAbstractItemModel::LayoutChangeHint hint)
 {
+    m_pendingDataChanged.clear();
```
vkrause: Are you sure this isn't too aggressive? IIRC the client keeps some data outside of the affected areas, which would potentially show outdated information if we drop the changes here, no?

A safer alternative might be to send all pending changes at this point.

pasnox: Hm, indeed it could be some issues with Qt 5.x and parents not empty.
I guess we want to send all (i would like to avoid sending obsolete data), or remove the entries having a parent in parents.
Will check what make best.

pasnox: Fixed

plugins/signalmonitor/signalhistorydelegate.cpp
pasnox: Done

plugins/signalmonitor/signalhistorymodel.h
vkrause: The extra column costs 70px here, I think that's too much. Also, "M." is not exactly an easily understandable title, even I have no idea what "M." stands for :) So, if you insist on keeping the column, can we at least shrink it to the point it hardly needs more space than the checkbox it contains?

plugins/signalmonitor/signalhistorydelegate.cpp
vkrause: The FPS slider is still there though?

pasnox: Yes, would you prefer it to be removed ?

plugins/signalmonitor/signalhistorymodel.h
pasnox: Will do

pasnox: The very minimum width for this column has to be check indicator width + sort indicator width + spacing in between, giving approximatively 40pixels here.
Ig this comes unacceptable for you, then the only option is to remove such column :(

PR #378, opened 2017-06-28
Smart host Edit

chsterz:
This change should replace the old launcher page with its Two-Part Url Input (url + port).
Its designed to be:
1. Versatile in accepting Urls (Localsockets, IPv4, all IPv6 formats, IPv4 embedded in IPv6)
2. Interactive (changing textcolor to red if address not accepted so far)
3. Forgiving for recoverable mistakes (forgetting 'local://' or 'tcp://' or the default port)
4. Suggestive (warning when default port was chosen automatically, alert when file is not socket)


launcher/ui/connectpage.cpp
```diff
@@ -34,19 +34,34 @@
...
 ConnectPage::ConnectPage(QWidget *parent)
     : QWidget(parent)
     , ui(new Ui::ConnectPage)
+    , m_localPrefix("local://")
+    , m_tcpPrefix("tcp://")
 {
     ui->setupUi(this);
-
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    m_implicitPortWarningSign = new QAction(QIcon(":/launcher/warning.png"), "No port given, will use 11732", this);
```
vkrause: missing tr() around the user-facing message

launcher/ui/connectpage.cpp
```diff
@@ -34,19 +34,34 @@
...
+    m_fileIsNotASocketWarning = new QAction(qApp->style()->standardIcon(QStyle::SP_MessageBoxCritical)
+                                            , "File is not a socket"
```
vkrause: same here

launcher/ui/connectpage.cpp
```diff
@@ -55,9 +70,178 @@ ConnectPage::ConnectPage(QWidget *parent)
...
+
+    QRegExp hostNameFormat("^([a-z,A-Z][a-z,A-Z,0-9,\\-\\.]+[a-z,A-Z,0-9])$");
+    if (hostNameFormat.exactMatch(stillToParse))
+        handleHostName(stillToParse);
+
+    // if we came down here and theres nothing more to parse, we are good
+    // set text back to black again
+    if (correctSoFar && stillToParse.isEmpty()) {
+        m_valid = true;
+        palette.setColor(QPalette::Text,Qt::black);
```
vkrause: Hardcoded black color looks problematic for dark color schemes. Can we use the text color from the system palette instead?

launcher/ui/connectpage.cpp
```diff
@@ -34,19 +34,34 @@
...
+#include <sys/stat.h>
+#endif
+
 using namespace GammaRay;
 
 ConnectPage::ConnectPage(QWidget *parent)
     : QWidget(parent)
     , ui(new Ui::ConnectPage)
+    , m_localPrefix("local://")
+    , m_tcpPrefix("tcp://")
```
vkrause: Those two don't need to be non-static member variables I think, they seem to be used as constants.

launcher/ui/connectpage.cpp
```diff
@@ -55,9 +70,178 @@ ConnectPage::ConnectPage(QWidget *parent)
...
+{
+    if (hostInfo.error() != QHostInfo::NoError)
+        return;
+
+    if(hostInfo.addresses().empty())
+        return;
+
+    m_valid = true;
+    QPalette palette;
+    palette.setColor(QPalette::Text,Qt::black);
```
vkrause: see above

PR #385, opened 2017-07-13
Fix missing g++ in cmake toolchain file for imx61

chgans:
Simple change to fix building for IMX61, without this change cmake picks up /usr/bin/c++ as the C++ cross-compiler.


chsterz:
Thanks for this contribution, Christian :)

PS: if you are using yocto, sourcing the yocto-provided environment-script usually helps cmake in finding all the necessary tools.
should be called something like: environment-setup-cortexa9hf-neon-poky-linux-gnueabi and is located at the base of the SDK.
Theres a generic Toolchain-Yocto.cmake for this case also.

chgans:
Thanks for the tip but i'm not using yocto.

PR #389, opened 2017-07-19
Handle byte array property editing like we do for string values

TheOneRing:
Task-Id: KDEND-192

### DIFF
--- a/ui/propertyeditor/propertyeditorfactory.cpp
+++ b/ui/propertyeditor/propertyeditorfactory.cpp
@@ -45,6 +45,7 @@ PropertyEditorFactory::PropertyEditorFactory()
     initBuiltInTypes();
 
     addEditor(QVariant::Color, new QStandardItemEditorCreator<PropertyColorEditor>());
+    addEditor(QVariant::ByteArray, new QStandardItemEditorCreator<PropertyByteArrayEditor>());
     addEditor(QVariant::Font, new QStandardItemEditorCreator<PropertyFontEditor>());
     addEditor(QVariant::Palette, new QStandardItemEditorCreator<PropertyPaletteEditor>());
     addEditor(QVariant::Point, new QStandardItemEditorCreator<PropertyPointEditor>());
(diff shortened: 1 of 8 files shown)


ui/propertyeditor/propertyeditorfactory.cpp
```diff
@@ -45,6 +45,7 @@ PropertyEditorFactory::PropertyEditorFactory()
     initBuiltInTypes();
 
     addEditor(QVariant::Color, new QStandardItemEditorCreator<PropertyColorEditor>());
+    addEditor(QVariant::ByteArray, new QStandardItemEditorCreator<PropertyTextEditor>());
```
pasnox: I did check nothing but is not the editor would change the QByteArray type to a QString here ?

TheOneRing: I'd expect the QVariant magic to get the correct type.
I tested it with fragmentShader and the changes where applied as expected.

pasnox: Hm what if the QString is not ascii encoded ? does it keep the correct chars like éàç ?

ui/propertyeditor/propertytexteditor.cpp
```diff
@@ -52,39 +52,42 @@ QString PropertyTextEditorDialog::text() const
...
-    // The user already pressed Apply, don't force her/him to do again
-    QKeyEvent event(QEvent::KeyPress, Qt::Key_Enter, Qt::NoModifier);
-    QApplication::sendEvent(this, &event);
 }
 
-void PropertyTextEditor::edit()
+void PropertyByteArrayEditor::edit()
 {
-    PropertyTextEditorDialog dlg(text(), this);
+    PropertyTextEditorDialog dlg(value().toByteArray(), this);
```
pasnox: Let be clear we assume all utf8 by default and for Qt4 support, let use QString::fromUtf8()

PR #393, opened 2017-07-21
Fix picking

chsterz:


plugins/quickinspector/quickinspector.cpp
```diff
@@ -243,9 +243,9 @@ static QString qsgTextureWrapModeToString(QSGTexture::WrapMode wrapMode)
 
 static bool isGoodCandidateItem(QQuickItem *item)
 {
-    if (!item->isVisible() || item->z() < 0 || qFuzzyCompare(item->opacity() + 1.0, qreal(1.0)) ||
```
pasnox: Can't we make those properties instead of removing them ?
Those were asked features by kdabians on different projects.

milianw: but z < 0 doesn't mean it's invisible, so why should we exclude it? it could still be visible if all other siblings have a z even smaller

and pure quick item check also sounds bogus, as this fails with subclasses. so overall the change by christoph sounds fine to me, but the commit message needs to be improved to explain why these changes are done

pasnox: The problem is that depending the project those kind of items are assumed *background* / *not interesting* items while picking. I also find them wrong because for me looks like wrong written code. So let remove and we will see if somebody complains or not.

milianw: Yep, it's ugly code - I completely agree. But GammaRay tends to be used in the face of ugly/broken code and needs to cope with that.

tests/quickinspectorpickingtest.cpp
```diff
@@ -0,0 +1,208 @@
...
+        Paths::setRelativeRootPath(GAMMARAY_INVERSE_BIN_DIR);
+        qputenv("GAMMARAY_ProbePath", Paths::probePath(GAMMARAY_PROBE_ABI).toUtf8());
+        qputenv("GAMMARAY_ServerAddress", GAMMARAY_DEFAULT_LOCAL_TCP_URL);
+        Hooks::installHooks();
+        Probe::startupHookReceived();
+        new ProbeCreator(ProbeCreator::Create);
+        QTest::qWait(1); // event loop re-entry
+    }
+
+    void triggerSceneChange()
```
vkrause: seems unused

tests/quickinspectorpickingtest.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        auto sgSelectionModel = ObjectBroker::selectionModel(sgModel);
+        QVERIFY(sgModel);
+        QSignalSpy sgSpy(sgSelectionModel, SIGNAL(selectionChanged(QItemSelection,QItemSelection)));
+        QVERIFY(sgSpy.isValid());
+
+        // auto center-click is broken before https://codereview.qt-project.org/141085/
+        QTest::mouseClick(view, Qt::LeftButton, Qt::ShiftModifier | Qt::ControlModifier,
+                          QPoint(view->width()/2, view->height()/2));
+        QTest::qWait(20);
```
vkrause: hardcoded timing are brittle on the CI, maybe better use QVERIFY(xxxSpy.wait())?

tests/quickinspectorpickingtest.cpp
```diff
@@ -0,0 +1,208 @@
...
+        createProbe();
+
+        // we need one view for the plugin to activate, otherwise the model will not be available
+        view = new QQuickView;
+        view->setResizeMode(QQuickView::SizeViewToRootObject);
+        QTest::qWait(1); // event loop re-entry
+
+        itemModel = ObjectBroker::model(QStringLiteral("com.kdab.GammaRay.QuickItemModel"));
+        QVERIFY(itemModel);
+        new ModelTest(itemModel, view);
```
vkrause: The model testing code here and below looks like a copy/paste leftover we probably don't need here?

tests/quickinspectorpickingtest.cpp
```diff
@@ -0,0 +1,208 @@
...
+    {
+        QTest::addColumn<QString>("qmlFile", nullptr);
+        QTest::addColumn<QString>("pickedObjectId", nullptr);
+
+        QTest::newRow("Vanilla Rect-Clicking") << "qrc:/manual/picking/stackedrects.qml" << "bluerect";
+        QTest::newRow("Negative z-order") << "qrc:/manual/picking/negativezordering.qml" << "greenrect";
+        QTest::newRow("Invisible overlay") << "qrc:/manual/picking/invisibleoverlay.qml" << "redrect";
+        QTest::newRow("Opacity:0 overlay") << "qrc:/manual/picking/opacityzerooverlay.qml" << "yellowrect";
+        QTest::newRow("Loader") << "qrc:/manual/picking/loader.qml" << "bluerect";
+        //Does not Work
```
vkrause: probably should have a corresponding QEXPECT_FAIL to not annoy the CI?

plugins/quickinspector/quickinspector.cpp
vkrause: I also think the new picking logic is an improvement regarding correctness.

PR #397, opened 2017-07-25
Add the GammaRay.ico to the launchers

TheOneRing:


cmake/GammaRayMacrosInternal.cmake
```diff
@@ -106,6 +108,13 @@ macro(gammaray_embed_info_plist _target _plist)
   endif()
 endmacro()
 
+macro(gammaray_add_win_icon _sources)
+    if (WIN32)
+        configure_file(${_gammaray_macros_internal_location}/gammaray.rc.cmake gammaray.rc)
+        LIST(APPEND ${_sources} ${CMAKE_CURRENT_BINARY_DIR}/gammaray.rc)
```
krf: Does this actually propagate? I think it'll create a new variable.

Don't you need `set(... PARENT_SCOPE)`?

TheOneRing: I guess as it is a macro I don't need to put it in the parent scope.
Or something like that
.

krf: Ah, right, sorry :)

PR #398, opened 2017-07-25
Cache QMetaProperty lookup in QuickOverlay::initFromItem

TheOneRing:


plugins/quickinspector/quickoverlay.cpp
```diff
@@ -50,6 +50,50 @@ extern Q_GUI_EXPORT QImage qt_gl_read_framebuffer(const QSize &size,
...
 
 namespace GammaRay {
+
+class QQuickItemPropertyCache {
+public:
+    static const QQuickItemPropertyCache &getPropertyCache(QQuickItem *item)
+    {
+        static QHash<QByteArray, QQuickItemPropertyCache> s_cache;
+        const auto meta = item->metaObject();
+        const auto className = QByteArray(meta->className());
```
pasnox: QByteArray(const char*) make a deep copy of the data.
I would use QHash<QLatin1String, QQuickItemPropertyCache> s_cache and QLatin1String(mo->className()).
And track item deletion to remove it from the hash.

TheOneRing: Hm I'll try that, but how should something be removed from the hash?
The meta classes should always exist?

pasnox: You can watch for QObject::destroyed (not sure you can still read the instance metaobject though, would need to check) and remove the corresponding classname cache.
The cache would be recreated later by another same class call and as items usually don't get deleted often should not make any difference.

TheOneRing: Connection this the object doesn't sound like the best idea.
Besides, the class name probably persists over the lifetime of the qobject?

pasnox: I have not idea if a QMetaObject instance classname data is still valid after the object destruction.
You would have to read the (generated) code for knowing that.
If it's still valid then of course you don't have to track objects for deletion.

TheOneRing: When you look at a moc file, the data is part of a static struct on the stack.

plugins/quickinspector/quickoverlay.cpp
```diff
@@ -50,6 +50,50 @@ extern Q_GUI_EXPORT QImage qt_gl_read_framebuffer(const QSize &size,
 QT_END_NAMESPACE
 
 namespace GammaRay {
+
+class QQuickItemPropertyCache {
+public:
+    static const QQuickItemPropertyCache &getPropertyCache(QQuickItem *item)
+    {
+        // className() is generated by moc and static -> we can use the pointer as key
```
vkrause: Not true for dynamic types as we encounter them in QML :) But more importantly, why do we need to look up properties per type at all? Are there really types overriding the properties in QQuickItem, or could we just look up the properties on that once and be done with it?

TheOneRing: As I understood @pasnox not all classes provide margins therefore we have to test for the class type.

Is any property cached a dynamic property?

vkrause: Ok, performance numbers showed that this approach is probably fast enough, and has the benefit of supporting random other types with similar property semantics too. I think we can use QMetaObject* as hash key directly then, rather than the className() pointer, to make this a bit easier to understand?

plugins/quickinspector/quickoverlay.cpp
```diff
@@ -45,6 +45,49 @@
 #include <cmath>
 
 namespace GammaRay {
+
+class QQuickItemPropertyCache {
+public:
+    static const QQuickItemPropertyCache &getPropertyCache(QQuickItem *item)
+    {
+        static QHash<quintptr, QQuickItemPropertyCache> s_cache;
```
vkrause: Can't we actually use QMetaObject* here now, instead of quintptr? That would avoid all the ugly reinterpret_casts below, and make it much clearer what the key actually is.

PR #400, opened 2017-07-28
Add quick remote view unit test

pasnox:
Also added base classes for unit testing.

### DIFF
--- a/core/remoteviewserver.cpp
+++ b/core/remoteviewserver.cpp
@@ -50,6 +50,7 @@ RemoteViewServer::RemoteViewServer(const QString &name, QObject *parent)
     , m_clientActive(false)
     , m_sourceChanged(false)
     , m_clientReady(true)
+    , m_grabberReady(true)
     , m_pendingReset(false)
     , m_pendingCompleteFrame(false)
 {
@@ -89,6 +90,14 @@ bool RemoteViewServer::isActive() const
     return m_clientActive;
 }
 
+void RemoteViewServer::setGrabberReady(bool ready)
+{
+    if (ready == m_grabberReady)
+        return;
+    m_grabberReady = ready;
+    checkRequestUpdate();
+}
+
 void RemoteViewServer::sendFrame(const RemoteViewFrame &frame)
 {
     m_clientReady = false;
@@ -125,7 +134,8 @@ void RemoteViewServer::clientViewUpdated()
 
 void RemoteViewServer::checkRequestUpdate()
 {
-    if (isActive() && !m_updateTimer->isActive() && m_clientReady && m_sourceChanged)
+    if (isActive() && !m_updateTimer->isActive() &&
+            m_clientReady && m_grabberReady && m_sourceChanged)
         m_updateTimer->start();
 }
 
--- a/plugins/quickinspector/quickoverlay.cpp
+++ b/plugins/quickinspector/quickoverlay.cpp
@@ -197,8 +197,6 @@ QuickOverlay::QuickOverlay()
     Q_ASSERT(m_sceneGrabbed.methodIndex() != -1);
     m_sceneChanged = mo->method(mo->indexOfSignal(QMetaObject::normalizedSignature("sceneChanged()")));
     Q_ASSERT(m_sceneChanged.methodIndex() != -1);
-    m_setIsGrabbingMode = mo->method(mo->indexOfSlot(QMetaObject::normalizedSignature("setIsGrabbingMode(bool)")));
-    Q_ASSERT(m_setIsGrabbingMode.methodIndex() != -1);
 
     qRegisterMetaType<GrabbedFrame>();
 }
@@ -306,16 +304,6 @@ void QuickOverlay::placeOn(const ItemOrLayoutFacade &item)
     updateOverlay();
 }
 
-void QuickOverlay::setIsGrabbingMode(bool isGrabbingMode)
-{
-    if (m_isGrabbingMode == isGrabbingMode)
-        return;
-
-    m_isGrabbingMode = isGrabbingMode;
-    if (m_isGrabbingMode)
-        updateOverlay();
-}
-
 QuickItemGeometry QuickOverlay::initFromItem(QQuickItem *item) const
 {
     QuickItemGeometry itemGeometry;
@@ -398,6 +386,22 @@ QuickItemGeometry QuickOverlay::initFromItem(QQuickItem *item) const
     return itemGeometry;
 }
 
+void QuickOverlay::setGrabbingMode(bool isGrabbingMode, const QRectF &userViewport)
+{
+    QMutexLocker locker(&m_mutex);
+
+    if (m_isGrabbingMode == isGrabbingMode)
+        return;
+
+    m_isGrabbingMode = isGrabbingMode;
+    m_userViewport = userViewport;
+
+    emit grabberReadyChanged(!m_isGrabbingMode);
+
+    if (m_isGrabbingMode)
+        updateOverlay();
+}
+
 void QuickOverlay::windowAfterSynchronizing()
 {
     // We are in the rendering thread at this point
@@ -407,10 +411,12 @@ void QuickOverlay::windowAfterSynchronizing()
 
 void QuickOverlay::windowAfterRendering()
 {
+    QMutexLocker locker(&m_mutex);
+
     // We are in the rendering thread at this point
     // And the gui thread is NOT locked
-
     Q_ASSERT(QOpenGLContext::currentContext() == m_window->openglContext());
+
     if (m_isGrabbingMode) {
         const auto window = QRectF(QPoint(0,0), m_renderInfo.windowSize);
         const auto intersect = m_userViewport.isValid() ? window.intersected(m_userViewport) : window ;
@@ -437,9 +443,9 @@ void QuickOverlay::windowAfterRendering()
         m_grabbedFrame.transform.translate(intersect.x() * m_renderInfo.dpr , -intersect.y() * m_renderInfo.dpr - h);
         m_grabbedFrame.image.setDevicePixelRatio(m_renderInfo.dpr);
 
-        if (!m_grabbedFrame.image.isNull()) {
-            m_sceneGrabbed.invoke(this, Qt::QueuedConnection, Q_ARG(GammaRay::GrabbedFrame, m_grabbedFrame));
-        }
+        // Let emit the signal even if our image is possibly null, this way we make perfect ping/pong
+        // reuests making it easier to unit test.
+        m_sceneGrabbed.invoke(this, Qt::QueuedConnection, Q_ARG(GammaRay::GrabbedFrame, m_grabbedFrame));
     }
 
     drawDecorations();
@@ -447,7 +453,8 @@ void QuickOverlay::windowAfterRendering()
     m_window->resetOpenGLState();
 
     if (m_isGrabbingMode) {
-        setIsGrabbingMode(false);
+        locker.unlock();
+        setGrabbingMode(false, QRectF());
     } else {
         m_sceneChanged.invoke(this, Qt::QueuedConnection);
     }
@@ -606,9 +613,5 @@ void QuickOverlay::disconnectTopItemChanges(QQuickItem *item)
 
 void QuickOverlay::requestGrabWindow(const QRectF &userViewport)
 {
-    if (m_isGrabbingMode)
-        return;
-
-    m_userViewport = userViewport;
-    m_setIsGrabbingMode.invoke(this, Qt::QueuedConnection, Q_ARG(bool, true));
+    setGrabbingMode(true, userViewport);
 }
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -421,6 +421,13 @@ if(Qt5Core_FOUND AND NOT Qt5Core_VERSION_MINOR LESS 4) # requires QHooks
         )
         target_link_libraries(quickinspectortest gammaray_core gammaray_quickinspector_shared Qt5::Quick)
 
+        gammaray_add_quick_test(quickinspectortest2
+            quickinspectortest2.cpp
+            quickinspectortest.qrc
+            $<TARGET_OBJECTS:modeltestobj>
+            )
+        target_link_libraries(quickinspectortest2 gammaray_core gammaray_quickinspector_shared Qt5::Quick)
+
         gammaray_add_quick_test(quickinspectorpickingtest
             quickinspectorpickingtest.cpp
             quickinspectortest.qrc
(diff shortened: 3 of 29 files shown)


tests/actiontest.cpp
```diff
@@ -169,9 +152,9 @@ private slots:
         QCOMPARE(a1->isChecked(), false);
         QCOMPARE(changeSpy.size(), 2);
     }
-
 };
 
 QTEST_MAIN(ActionTest)
 
+#include "moc_baseprobetest.cpp"
```
vkrause: I very much like the idea. Can we compile this only once like we do for the probe object, and add this via CMake though, especially since there is now the gammaray_add_probe_test macro? Would avoid the extra moc include too.

core/remoteviewserver.cpp
```diff
@@ -228,5 +228,6 @@ void RemoteViewServer::clientConnectedChanged(bool connected)
 void RemoteViewServer::requestUpdateTimeout()
 {
     emit requestUpdate();
+    emit frameRequested();
```
vkrause: Why the extra signal, isn't requestUpdate() emitted at exactly the same time? Testing that would even be better, as that's what will be what the using code will follow.

tests/quickinspectortest.cpp
```diff
@@ -225,6 +226,93 @@ private slots:
...
+        QVERIFY(rootItem);
+        Probe::instance()->selectObject(rootItem, QPoint());
+
+        connect(remoteView, SIGNAL(frameUpdated(GammaRay::RemoteViewFrame)), remoteView, SLOT(clientViewUpdated()));
+
+        // RemoteViewServer throttle the requests with an interval of ~16ms by default
+        qreal throttlerInterval = 100.0;
+        // Our animation properties
+        qreal animationInterval = 0.0;
+        qreal animationDuration = 1400.0;
```
vkrause: Testing for real frame timings looks extremely brittle, wouldn't this randomly break on high CI load?

core/remoteviewserver.cpp
```diff
@@ -54,7 +54,9 @@ RemoteViewServer::RemoteViewServer(const QString &name, QObject *parent)
                                                     name), this, "clientConnectedChanged");
 
     m_updateTimer->setSingleShot(true);
-    m_updateTimer->setInterval(10);
+    // Qml render @ ~60fps which is approximatively 16.66ms, not interesting to throttle under that value
+    // because we would probably resend the same frame again.
```
vkrause: That is actually wrong, it does not resend in that case. It also checks if the source actually changed. At least that's what it is supposed to be doing, would be nice to test as well I guess. Ie. if you have a static scene, it should only send you one frame, until you tell it the scene has changed.

common/remoteviewinterface.h
```diff
@@ -76,6 +76,8 @@ public slots:
 
     virtual void setViewActive(bool active) = 0;
 
+    virtual void setUpdateInterval(int ms) = 0;
```
vkrause: This is not something anyone externally should control, it's purely an implementation detail to reduce event loop load. Can we get away without having this somehow?

plugins/quickinspector/quickoverlay.cpp
```diff
@@ -437,16 +439,16 @@ void QuickOverlay::windowAfterRendering()
         m_grabbedFrame.transform.translate(intersect.x() * m_renderInfo.dpr , -intersect.y() * m_renderInfo.dpr - h);
         m_grabbedFrame.image.setDevicePixelRatio(m_renderInfo.dpr);
 
-        if (!m_grabbedFrame.image.isNull()) {
-            m_sceneGrabbed.invoke(this, Qt::QueuedConnection, Q_ARG(GammaRay::GrabbedFrame, m_grabbedFrame));
-        }
+        Q_ASSERT(!m_grabbedFrame.image.isNull());
```
vkrause: Sure this is always the case? Can w and h be 0 above?

plugins/quickinspector/quickoverlay.cpp
```diff
@@ -437,16 +439,16 @@ void QuickOverlay::windowAfterRendering()
...
+        m_sceneGrabbed.invoke(this, Qt::QueuedConnection, Q_ARG(GammaRay::GrabbedFrame, m_grabbedFrame));
     }
 
     drawDecorations();
 
     m_window->resetOpenGLState();
 
     if (m_isGrabbingMode) {
+        locker.unlock();
         setIsGrabbingMode(false);
```
vkrause: Seeing this pattern also below, wouldn't it be better to assume the lock being held in setIsGrabbingMode? So the value doesn't change between the check here, and the call to it?

tests/actiontest.cpp
pasnox: I think no because there is a macro inside it meaning it can be built 2 differents way.
Or I would have to dissociate both *probe* ways as different classes, then probably I can do that.

vkrause: Is that distinction actually necessary? or is that just because one test was written somewhat differently for historic reasons?

pasnox: Looks like we can merge the code path and go that way. Will do.

core/remoteviewserver.cpp
pasnox: Yep, though the signal is not part of the interface it's why I added a new one.
In the end I can indeed use this one using Qt 4 style connection.
Fixed

core/remoteviewserver.cpp
pasnox: It was in the expectation that we would force request it, in this case unless for Qt 3D scene, we would not have a changed result frame, will change the comment.
I will add a test for that as well.

common/remoteviewinterface.h
pasnox: Fixed

plugins/quickinspector/quickoverlay.cpp
pasnox: Fixed

plugins/quickinspector/quickoverlay.cpp
pasnox: Fixed

tests/actiontest.cpp
pasnox: Fixed

core/remoteviewserver.cpp
pasnox: Fixed

tests/quickinspectortest.cpp
pasnox: Fixed

plugins/quickinspector/quickoverlay.cpp
```diff
@@ -437,17 +441,18 @@ void QuickOverlay::windowAfterRendering()
         m_grabbedFrame.transform.translate(intersect.x() * m_renderInfo.dpr , -intersect.y() * m_renderInfo.dpr - h);
         m_grabbedFrame.image.setDevicePixelRatio(m_renderInfo.dpr);
 
-        if (!m_grabbedFrame.image.isNull()) {
-            m_sceneGrabbed.invoke(this, Qt::QueuedConnection, Q_ARG(GammaRay::GrabbedFrame, m_grabbedFrame));
-        }
+        // Let emit the signal even if our image is possibly null, this way we make perfect ping/pong
+        // reuests making it easier to unit test.
```
vkrause: Easier to unit test is not the design criteria here ;-) Can the client handle empty frames? How often does  this happen, ie. is this causing unreasonable performance costs?

tests/CMakeLists.txt
```diff
@@ -13,20 +14,76 @@ else()
...
 endif()
 
+# create intermediate base_test_obj object files which unit tests can use
+add_library(base_test_obj OBJECT basetest.h)
+set_target_properties(base_test_obj PROPERTIES POSITION_INDEPENDENT_CODE ON)
+target_compile_definitions(base_test_obj PRIVATE ${Qt5Core_DEFINITIONS} ${Qt5Test_DEFINITIONS}) # for -DQT_NAMESPACE=...
+target_include_directories(base_test_obj SYSTEM PRIVATE
+    ${Qt5Core_INCLUDE_DIRS}
+    ${Qt5Test_INCLUDE_DIRS}
+)
```
vkrause: This looks like it might not build with Qt4?

tests/CMakeLists.txt
```diff
@@ -13,20 +14,76 @@ else()
...
+set_target_properties(base_probe_test_obj PROPERTIES POSITION_INDEPENDENT_CODE ON)
+target_compile_definitions(base_probe_test_obj PRIVATE ${Qt5Core_DEFINITIONS} ${Qt5Test_DEFINITIONS}) # for -DQT_NAMESPACE=...
+target_include_directories(base_probe_test_obj PUBLIC
+    $<TARGET_PROPERTY:gammaray_core,INTERFACE_INCLUDE_DIRECTORIES>
+    $<TARGET_PROPERTY:gammaray_common,INTERFACE_INCLUDE_DIRECTORIES>
+)
+target_include_directories(base_probe_test_obj SYSTEM PRIVATE
+    ${Qt5Core_INCLUDE_DIRS}
+    ${Qt5Test_INCLUDE_DIRS}
+)
```
vkrause: same here

tests/basetest.h
```diff
@@ -0,0 +1,131 @@
...
+
+#include <config-gammaray.h>
+
+#include <QObject>
+#include <QSignalSpy>
+#include <QAbstractItemModel>
+
+#include <QtTest/qtest.h>
+
+class BaseTest : public QObject
```
vkrause: Good idea to factor out those utilities. They don't need to be in the inheritance hierarchy, right? So we could have them in a separate utility namespace for example, would make it easier to combine this with the other base classes.

tests/quickinspectortest.cpp
```diff
@@ -258,18 +203,123 @@ private slots:
...
+            rootItem->setProperty("interval", qRound(animationInterval));
+            rootItem->setProperty("animated", true);
+            QTest::qWait(qRound(animationDuration));
+
+            // Wait to process pending requests...
+            rootItem->setProperty("animated", false);
+            QTest::qWait(qRound(animationDuration));
+
+            QVERIFY(requestedSpy.count() <= qRound(maxPossibleThrottledRequests * 1.05));
+            QVERIFY(updatedSpy.count() <= qRound(maxPossibleThrottledRequests * 1.05));
```
vkrause: We never call clientViewUpdated() here, so shouldn't this be always 1 here, no matter how much the source actually changed? That would actually be a valid test-case, independent of what I suspect this code tried to test :)

common/remoteviewinterface.h
```diff
@@ -85,6 +85,8 @@ public slots:
     void reset();
     void elementsAtReceived(const GammaRay::ObjectIds &ids, int bestCandidate);
     void frameUpdated(const GammaRay::RemoteViewFrame &frame);
+    /// when server receive this signal, obtain a new frame and send it to the client
+    void requestUpdate();
```
pasnox: Hm, I can move it back to server again , that's indeed a performance penalty.
The issue was that I then can't do any Qt5 connect style to a lambda.
With Qt 4 connect style I would need a slot, being *checked* by the unit test framework.

core/remoteviewserver.cpp
```diff
@@ -57,7 +57,9 @@ RemoteViewServer::RemoteViewServer(const QString &name, QObject *parent)
                                                     name), this, "clientConnectedChanged");
 
     m_updateTimer->setSingleShot(true);
-    m_updateTimer->setInterval(10);
+    // Qml try to render @ ~60fps which is approximatively 16.66ms, not interesting to throttle
+    // under that value.
+    m_updateTimer->setInterval(qRound(1000.0 / 60.0));
```
pasnox: Fixed

core/remoteviewserver.cpp
```diff
@@ -91,7 +93,6 @@ bool RemoteViewServer::isActive() const
 
 void RemoteViewServer::sendFrame(const RemoteViewFrame &frame)
 {
-    m_clientReady = false;
```
pasnox: The reason was specified in the commit message as "It was possible to trigger a new update request if sourceChanged
was called in between requestUpdateTimeout and sendFrame."
The sourceChanged being directly connected to the QuickOverlay::sceneChanged() signal.
In rare case the signal is emitted in between the 2 quoted calls which trigger a non wanted update request.

plugins/quickinspector/quickoverlay.cpp
pasnox: Yes it can handle null images.
Technically I never had a case of null image, but I suppose it can arrive if the window root item has a size of 0x0 (yeah that's improbable case but still possible).
Does cost nothing as far as I can see.

core/remoteviewserver.cpp
vkrause: But this was correct, no? If the source changed after we grabbed the frame we eventually need to get a second frame. After all we want to show the correct state in the client.

pasnox: Hm I don't think so.
If we reach requestUpdateTimeout a QuickOverlay grab request is triggered.
Depending at which time the request is done, we can already be in locked state in QuickOverlay::windowAfterRendering which will emit a sceneChanged just before we can change the grabbing mode. This will then trigger a new update request while we did not even yet changed the QuickOverlay grabbing mode. As a result the next request timeout will try to request a frame grab but **could** do nothing because we **could** be in the process of capturing a frame already (ie, grabbing mode is true already).
Schematized by: "Request -> Request -> Grab" **instead of** "Request -> Grab -> Request -> Grab"
So for me if the client requested a frame but did not yet received it, I think a request update can't occurs server side as normally only a clientViewUpdated call could trigger a possible frame grab (assaying - client is ready).
As a consequence, requesting a frame update should mark client as non ready.
If my reasoning sound wrong to you, then probably we want to fix the double update request calls (ie, by introducing a quickoverlay grabbing state test in RemoteViewServer::checkRequestUpdate() so the new request will only occurs after the current grab finished).

pasnox: After writing all that it make it clear that the problem comes from the differences between the different grabbers... widget grabber is synchrone while the quick one is asynchrone now (used to be synchrone before but no longer). So we did not get that corner case before.

tests/quickinspectortest.cpp
pasnox: We do call it, from the upper connection.

plugins/quickinspector/quickoverlay.cpp
pasnox: Fixed

tests/CMakeLists.txt
pasnox: Fixed

tests/CMakeLists.txt
pasnox: Fixed

tests/basetest.h
pasnox: Fixed

tests/quickinspectortest.cpp
pasnox: Fixed

core/remoteviewserver.cpp
pasnox: Fixed

PR #404, opened 2017-08-07
Add screen shot support in the quick inspector

pasnox:
We can screen with and without client decorations.

Task-Id: KDEND-141


pasnox:
Fixed partial frame issue.

PR #405, opened 2017-08-07
Include MSVC versions in the probe ABI

TheOneRing:
Also include the debug version in for mingw builds.

### DIFF
--- a/cmake/GammaRayProbeABI.cmake
+++ b/cmake/GammaRayProbeABI.cmake
@@ -42,6 +42,23 @@ endif()
 # on Windows, the compiler also matters
 if(WIN32)
   set(GAMMARAY_PROBE_ABI "${GAMMARAY_PROBE_ABI}-${CMAKE_CXX_COMPILER_ID}")
+  if(MSVC)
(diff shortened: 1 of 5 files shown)


cmake/GammaRayProbeABI.cmake
```diff
@@ -42,6 +42,21 @@ endif()
 # on Windows, the compiler also matters
 if(WIN32)
   set(GAMMARAY_PROBE_ABI "${GAMMARAY_PROBE_ABI}-${CMAKE_CXX_COMPILER_ID}")
+  if(MSVC)
+      # based on FindBoost.cmake
+      if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19.10)
+        set(GAMMARAY_COMPILER_VERSION "140")
+      elseif (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19)
+        set(GAMMARAY_COMPILER_VERSION "140")
```
vkrause: Don't the first two conditions map to the same thing? Is that intended?

PR #411, opened 2017-08-09
Use architecture specific injector on Windows 

TheOneRing:


launcher/cli/main.cpp
```diff
@@ -154,6 +155,31 @@ static bool startLauncher()
...
+
+    auto arguments = qApp->arguments();
+    arguments.pop_front();
+    auto env = options.processEnvironment();
+    const auto prepend = [&env](const QString &key, const QString& val) {
+        const QString path = val + QLatin1Char(';') + env.value(key, QString());
+        env.insert(key,  path);
+    };
+    prepend(QLatin1String("Path"), options.workingDirectory());
+    prepend(QLatin1String("QT_PLUGIN_PATH"), options.workingDirectory());
```
TheOneRing: this might not always include all plugin locations ....

TheOneRing:
Currently I don't see another scenario where we would want to run an injector with a different abi, so I put the code into a ifdef.

launcher/cli/main.cpp
```diff
@@ -215,6 +243,8 @@ int main(int argc, char **argv)
         }
         if ((arg == QLatin1String("-p") || arg == QLatin1String("--pid")) && !args.isEmpty()) {
             options.setPid(args.takeFirst().toInt());
+            const auto qtCorePath = ProbeABIDetector().qtCoreForProcess(options.pid());
+            options.setWorkingDirectory(qtCorePath.mid(0, qtCorePath.lastIndexOf("\\")));
```
vkrause: this still looks windows-only?

TheOneRing: ah sorry ...

PR #416, opened 2017-08-16
Install the attach dialog widget and prepare it to be embeddable 

TheOneRing:
The other 2 commits are from https://github.com/KDAB/GammaRay/pull/415

### DIFF
--- a/launcher/ui/attachdialog.cpp
+++ b/launcher/ui/attachdialog.cpp
@@ -31,6 +31,9 @@
 #include "processfiltermodel.h"
 #include "processmodel.h"
 #include "probeabimodel.h"
+#include "processlist.h"
+
+#include "ui_attachdialog.h"
 
 #include <launcher/core/launchoptions.h>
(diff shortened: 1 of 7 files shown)


launcher/ui/attachdialog.cpp
```diff
@@ -66,56 +70,66 @@ AttachDialog::AttachDialog(QWidget *parent, Qt::WindowFlags f)
...
     setWindowIcon(QIcon(QStringLiteral(":/gammaray/GammaRay-128x128.png")));
 
-    ui.stackedWidget->setCurrentWidget(ui.loadingLabel);
+    ui->stackedWidget->setCurrentWidget(ui->loadingLabel);
     updateProcesses();
 }
 
+AttachDialog::~AttachDialog()
+{
+    delete ui;
```
vkrause: more elegant would be using std::unique_ptr, that would avoid the need for the explicit deletion

PR #419, opened 2017-08-18
RemoteView: ColorPipette

chsterz:
* Additional mode of interaction
* Shows RGBA values along a bigger square of the hovered color
* Copying of Colors is possible, but I havn't found an app to paste it to


plugins/quickinspector/textureextension/textureviewwidget.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include "textureviewwidget.h"
 
+#include <QMouseEvent>
```
vkrause: unusued?

ui/remoteviewwidget.cpp
```diff
@@ -1136,6 +1174,39 @@ void RemoteViewWidget::contextMenuEvent(QContextMenuEvent *event)
     case InputRedirection:
         QWidget::contextMenuEvent(event);
         break;
+        break;
```
vkrause: breaking twice? ;-)

ui/trailingcolorlabel.cpp
```diff
@@ -0,0 +1,64 @@
+#include "trailingcolorlabel.h"
```
vkrause: license header missing here and below

ui/trailingcolorlabel.cpp
```diff
@@ -0,0 +1,64 @@
...
+    m_pickedColor = qRgba(0,0,0,0);
+    setHidden(true);
+}
+
+void TrailingColorLabel::paintEvent(QPaintEvent *event)
+{
+    Q_UNUSED(event);
+    QPainter *p = new QPainter(this);
+    p->setPen(QPen(Qt::NoPen));
+    const auto palette = this->style()->standardPalette();
```
vkrause: just "palette()" should work here, it's a property of QWidget

ui/trailingcolorlabel.cpp
```diff
@@ -0,0 +1,64 @@
...
+    p->setBrush(palette.base());
+    p->drawRect(0,0,145,30);
```
vkrause: This, and the numbers below seem to assume a specific font size, which is likely not the same everywhere. Probably needs to depend on fontMetrics() instead.

PR #422, opened 2017-08-21
Fix overriding of IVI properties

mkrus:
When setting overridden values, there was no way to know if the values
needed to be pushed to the backend. Setter now returns true if the
values are overridden.

Also fixed copy constructor.

This depends on a qtivi change I53444099520f75cb48f334591104aca9eb18d41e


jpetersen:
Looks good to be merged

PR #424, opened 2017-08-23
Update property overrides to handle multiple frontend instances

mkrus:
Multiple frontend instances will talk to the same backend plugin
instances. Users will expect the model to show the values once
rather have duplicates for all the instances.

This creates an overrider object per backend instance and manages
the list of front end instances. Changes to values from the plugin
UI are propagated to all the instances (which is needed if overriding
is one, in which case values don’t make it to the backend).

Also fix crash on shutdown with models not being cleaned up properly.

Did some class renaming for clarity.

This depends on QtIVI change I2d756106c14536936451f61a502a13a1bb28e8dd

### DIFF
--- a/plugins/qtivi/qtiviobjectmodel.cpp
+++ b/plugins/qtivi/qtiviobjectmodel.cpp
@@ -63,34 +63,18 @@ static const quint64 PropertyCarrierIndex = ~quint64(0); // there is no QUINT64_
 
 using namespace GammaRay;
 
-QtIviObjectModel::QtIviObjectModel(Probe *probe)
-    : QAbstractItemModel(probe)
-{
-    // Known wrongly qualified types
-    qRegisterMetaType<QIviAbstractFeature::DiscoveryMode>();
-    qRegisterMetaType<QIviAbstractFeature::DiscoveryResult>();
-#if defined(QT_IVIMEDIA_LIB)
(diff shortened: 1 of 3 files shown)


plugins/qtivi/qtiviobjectmodel.cpp
```diff
@@ -446,43 +509,58 @@ void QtIviObjectModel::objectAdded(QObject *obj)
...
+        auto it = std::find_if(m_serviceCarriers.begin(), m_serviceCarriers.end(), [serviceId, zoneId](const ServiceZone &p) {
+            return p.m_service->serviceId() == serviceId && p.m_zone == zoneId;
+        });
+        if (it != m_serviceCarriers.end()) {
+            (*it).m_service->addCarrier(featureObj);
+        } else {
+            beginInsertRows(QModelIndex(), m_serviceCarriers.size(), m_serviceCarriers.size());
+            ServiceZone sz;
+            sz.m_zone = zoneId;
+            sz.m_service = IviOverriderPtr::make_shared(featureObj);
```
jpetersen: Does not compile for me, should be std::make_shared<IviOverrider>?

mkrus: looks like it yes, will update

PR #430, opened 2017-09-04
don't review - used to trigger CI to test integration

jangmarker:


jangmarker:
tests done, seems to work fine

PR #436, opened 2017-09-08
Mark injected processes by overaying their icon and chhanging their t…

TheOneRing:
…itle


plugins/guisupport/guisupport.cpp
```diff
@@ -374,6 +387,60 @@ void GuiSupport::registerVariantHandler()
...
+        return;
+    w->setTitle(w->title() + m_titleSuffix);
+}
+
+void GuiSupport::updateWindowIcon(QWindow *w)
+{
+    static QVector<QPixmap> gammarayIcons;
+    static const auto theme = QLatin1String("gammaray");
+    if (gammarayIcons.isEmpty()) {
+        gammarayIcons << QPixmap(QLatin1String(":/gammaray/images/gammaray-inject-16.png"))
```
TheOneRing: We don't use the x2 and x3 icons, should we?

PR #438, opened 2017-09-15
Timer top model status update

pasnox:
Add a way to update a timer state change while it did not triggered a timeout recently.

### DIFF
--- a/plugins/timertop/clienttimermodel.cpp
+++ b/plugins/timertop/clienttimermodel.cpp
@@ -31,6 +31,10 @@
 #include "timerinfo.h"
 #include "timermodel.h"
 
+#include <QApplication>
+#include <QFont>
+#include <QColor>
+
 using namespace GammaRay;
 
 ClientTimerModel::ClientTimerModel(QObject *parent)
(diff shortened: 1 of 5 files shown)


pasnox:
This covers both QTimer and raw timers.
I will check for QQmlTimer support now, probably using something like runningChanged signal as this class does not expose a timer id.

pasnox:
QQmlTimer support added

plugins/timertop/clienttimermodel.cpp
```diff
@@ -44,24 +48,60 @@ ClientTimerModel::~ClientTimerModel()
...
+            const TimerId::Type type = TimerId::Type(typeSibling.data(TimerModel::TimerTypeRole).toInt());
+            QFont font = QApplication::font("QAbstractItemView");
+            font.setStrikeOut(type == TimerId::InvalidType || state == TimerIdInfo::InvalidState);
+            return QVariant::fromValue(font);
+        } else if(role == Qt::BackgroundRole) {
+            const QModelIndex sibling = index.sibling(index.row(), TimerModel::ObjectNameColumn);
+            const TimerId::Type type = TimerId::Type(sibling.data(TimerModel::TimerTypeRole).toInt());
+            switch (type) {
+            case TimerId::InvalidType:
+                return QColor(255, 0, 0, 80);
```
vkrause: Do these hardcoded colors work with light and dark color schemes?

pasnox: Tested on both, looks ok.

pasnox:
@volker ?

PR #439, opened 2017-09-18
Add 'client only' build option

pasnox:
Task-Id: KDEND-225

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@
 #  Build only an additional probe configuration for an already existing launcher.
 #  Default=false
 #
+# -DGAMMARAY_CLIENT_ONLY_BUILD=[true|false]
+#  Build the client part only.
+#  Default=false
+#
 # -DGAMMARAY_ENABLE_GPL_ONLY_FEATURES=[true|false]
 #  Enable features only available under GPL license.
 #  Default=false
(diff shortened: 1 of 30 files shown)


CMakeLists.txt
```diff
@@ -170,14 +174,30 @@ set(GAMMARAY_BUILD_UI_DEFAULT ON)
...
 
+gammaray_option(
+    GAMMARAY_CLIENT_ONLY_BUILD
+    "Build the client part only."
+    OFF
+)
+
+if (GAMMARAY_CLIENT_ONLY_BUILD)
+    set(GAMMARAY_BUILD_UI_DEFAULT ON)
+    set(BUILD_TESTING OFF)
```
vkrause: Why disable testing in this case?

CMakeLists.txt
```diff
@@ -752,10 +772,14 @@ if(GAMMARAY_BUILD_UI AND (Qt5Widgets_FOUND OR QT_QTGUI_FOUND))
...
   endif()
   add_subdirectory(ui)
-  add_subdirectory(inprocessui)
+  if (NOT GAMMARAY_CLIENT_ONLY_BUILD)
+    add_subdirectory(inprocessui)
+  endif()
   if(NOT GAMMARAY_PROBE_ONLY_BUILD)
     add_subdirectory(client)
-    add_subdirectory(app)
+    if (NOT GAMMARAY_CLIENT_ONLY_BUILD)
```
vkrause: Isn't "app" the macos client/launcher combination and thus should be build in client-only mode too?

pasnox: Fixed

CMakeLists.txt
pasnox: Currently they caused lots of build issues in those, I will have a closer looks now that there is no much comments for the review.

pasnox: Fixed

PR #441, opened 2017-09-20
Fix icon overlay

TheOneRing:


plugins/guisupport/guisupport.cpp
```diff
@@ -486,13 +486,16 @@ void GuiSupport::updateWindowIcon(QWindow *w)
...
-                      << QPixmap(QLatin1String(":/gammaray/images/gammaray-inject-64.png"))
-                      << QPixmap(QLatin1String(":/gammaray/images/gammaray-inject-128.png"));
+        QIcon gammarayIcon(QLatin1String(":/gammaray/images/gammaray-inject-16.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-22.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-24.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-32.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-48.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-64.png"));
+        gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-128.png"));
+        foreach (QSize size, gammarayIcon.availableSizes()) {
```
pasnox: I would not store the pixmaps for one reason - they are dependent of the current screen dpi and moving the window to another screen with a different dpi would break rendering.
Just keep the static QIcon, and use QIcon::pixmap(window, ...) or directly QIcon::paint() when you want to update the window icon.
Also, in the event filter, I would also listen for the screen change event so you can update the icons correctly based on the screen dpi, see https://wiki.kdab.com/display/KDABTech/HiDPI for more informations.

plugins/guisupport/guisupport.cpp
```diff
@@ -503,11 +503,11 @@ void GuiSupport::updateWindowIcon(QWindow *w)
 
     QIcon newIcon;
     newIcon.setThemeName(theme);
-    foreach (const auto &pixmap, gammarayIcons) {
-        QPixmap pix(oldIcon.pixmap(oldIcon.actualSize(pixmap.size())));
+    foreach (const QSize size, gammarayIcon.availableSizes()) {
```
pasnox: missing & ref

plugins/guisupport/guisupport.cpp
```diff
@@ -503,11 +503,11 @@ void GuiSupport::updateWindowIcon(QWindow *w)
...
+        QPixmap pix(oldIcon.pixmap(oldIcon.actualSize(size)));
```
pasnox: Use the override taking a QWindow to render correctly on hidpi screens

PR #443, opened 2017-09-21
Unify application/window icon/title

pasnox:
this also fix high dpi issues.

### DIFF
--- a/plugins/guisupport/guisupport.cpp
+++ b/plugins/guisupport/guisupport.cpp
@@ -79,8 +79,7 @@ GuiSupport::GuiSupport(GammaRay::ProbeInterface *probe, QObject *parent)
     connect(m_probe->probe(), SIGNAL(objectCreated(QObject*)), SLOT(objectCreated(QObject*)));
 
     if (auto guiApp = qobject_cast<QGuiApplication*>(QCoreApplication::instance())) {
-        m_originalAppIcon = guiApp->windowIcon();
-        guiApp->setWindowIcon(createIcon(m_originalAppIcon));
+        updateWindowIcon();
 
         m_probe->installGlobalEventFilter(this);
         foreach (auto w , qGuiApp->topLevelWindows()) {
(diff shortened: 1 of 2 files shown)


plugins/guisupport/guisupport.h
```diff
@@ -53,14 +53,13 @@ class GuiSupport : public QObject
     void registerVariantHandler();
 #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
     void discoverObjects();
-    void updateWindowTitle(QWindow *w);
-    QIcon createIcon(const QIcon &oldIcon, QWindow *w=nullptr);
-    void updateWindowIcon(QWindow *w);
-    void restoreWindowIcon(QWindow *w);
+    void updateWindowTitle(QObject *o);
+    QIcon createIcon(const QIcon &oldIcon, QObject *o=nullptr);
```
TheOneRing: Hm either remove the default value or still use a QWindow and drop the cast etc as you use qApp directly anyhow .

plugins/guisupport/guisupport.cpp
```diff
@@ -79,8 +79,8 @@ GuiSupport::GuiSupport(GammaRay::ProbeInterface *probe, QObject *parent)
     connect(m_probe->probe(), SIGNAL(objectCreated(QObject*)), SLOT(objectCreated(QObject*)));
 
     if (auto guiApp = qobject_cast<QGuiApplication*>(QCoreApplication::instance())) {
-        m_originalAppIcon = guiApp->windowIcon();
-        guiApp->setWindowIcon(createIcon(m_originalAppIcon));
+        updateWindowIcon();
+        updateWindowTitle();
```
TheOneRing: no longer needed

plugins/guisupport/guisupport.cpp
```diff
@@ -481,9 +481,11 @@ void GuiSupport::registerVariantHandler()
 #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
 void GuiSupport::updateWindowTitle(QWindow *w)
 {
-    if (w->title().endsWith(m_titleSuffix))
-        return;
-    w->setTitle(w->title() + m_titleSuffix);
+    if (w) {
```
TheOneRing: don't set a default value for window

plugins/guisupport/guisupport.cpp
```diff
@@ -499,29 +501,33 @@ QIcon GuiSupport::createIcon(const QIcon &oldIcon, QWindow *w)
         gammarayIcon.addFile(QLatin1String(":/gammaray/images/gammaray-inject-128.png"));
     }
 
-    const QIcon &orgIcon = m_originalIcons[w];
-    if (!oldIcon.isNull() && (!orgIcon.isNull() || orgIcon.cacheKey() == oldIcon.cacheKey())) {
+    // Make sure to override the window icon with the app icon if it was initialized with it
+    if (w) {
+        const auto it = m_originalIcons.constFind(qApp);
+        const QIcon appIcon = it != m_originalIcons.constEnd() ? it.value() : qApp->windowIcon();
+        if (oldIcon.cacheKey() == appIcon.cacheKey() || oldIcon.cacheKey() == qApp->windowIcon().cacheKey()) {
```
TheOneRing: Why do you return when the oldIcon is the same as the app icon?
Wouldn't that be exactly the case where we want to change the icon?

plugins/guisupport/guisupport.cpp
```diff
@@ -499,29 +501,33 @@ QIcon GuiSupport::createIcon(const QIcon &oldIcon, QWindow *w)
...
+            return qApp->windowIcon();
+        }
+    }
+
+    const auto it = m_originalIcons.constFind(w ? (QObject *)w : (QObject *)qApp);
+    if (!oldIcon.isNull() && (it != m_originalIcons.constEnd() && (!it.value().isNull() || it.value().cacheKey() == oldIcon.cacheKey()))) {
```
TheOneRing: we now can directly use it != m_originalIcons.constEnd() istead of isNull.
isNull was used to test whether the default constructor was used.

plugins/guisupport/guisupport.cpp
pasnox: For window icons implicitly set from QApplication::windowIcon() there is Qt implementation issues in QWindow, see QTBUG-63340.
Therefore we would not want to over-overlay implicit icons which are already marked by the overlay.

plugins/guisupport/guisupport.cpp
TheOneRing: it doesn't make much sense to call the function without setting a window.
So instead of checking for w != null just don't pass a default parameter.

pasnox: Fixed

PR #449, opened 2017-10-05
Implement aboutToDetach and restore Icons and title on detach

TheOneRing:


TheOneRing:
Hm looks like it can happe that the probe is loaded after it was detached?
``````
[7072] ~Probe
[7072] ~GuiSupport
[7072] gammaray_probe_inject()
[7072] gammaray_probe_inject()
[7072] gammaray_probe_inject()
[7072] gammaray_probe_inject()
[7072] Comparators already registered for type GammaRay::ObjectId
[7072] Comparators already registered for type GammaRay::ObjectIds
[7072] gammaray_probe_inject()
[7072] Failed to connect to launcher, can't receive probe settings! "QLocalSocket::connectToServer: Invalid name"
[11568] Suspending

TheOneRing:
The reinject issue was fixed in https://github.com/KDAB/GammaRay/pull/451 .

Regarding the build issues, has moc problems with slots in a ifdef?

PR #450, opened 2017-10-06
Integrate QtIVI fixes from 5.10

mkrus:
fix is all the changes that were made on 5.10 for qtivi related changes...


vkrause:
Is that for 2.8 or master?

mkrus:
right, was supposed to be 2.8. Will update

PR #461, opened 2017-10-19
Allow to reattach on Windows

TheOneRing:


core/probe.h
```diff
@@ -100,6 +104,13 @@ class GAMMARAY_CORE_EXPORT Probe : public QObject, public ProbeInterface
      */
     static bool isInitialized();
 
+#ifdef Q_OS_WIN
```
TheOneRing: I think I might be able to simplify things by directly unload after I called inject?

TheOneRing: hm no, that would crash

PR #462, opened 2017-10-20
Added hints to find qdoc and qhelpgenerator

sgaist:
Depending on how Qt is installed and CMake configured, a system
qdoc can be used in place of the version provided by the Qt hinted
to at configuration time. This patch fixes that by adding a hint
entry to find_program so it starts in the same folder as the Qt
configured.


krf:
Thanks for the patch!

PR #463, opened 2017-10-23
This disable new (5.10) qtivi to keep Gammaray building

mkrus:
Work on qtivi integration is happening on 5.10


krf:
:+1: -- other opinions?

PR #476, opened 2017-11-28
Fix probe loading with default plugin layout

TheOneRing:
The probe name and the loader name clashed with the default
naming as the probe finder was basically looking for 'gammaray_probe*'


milianw:
Does not fix the debug build for me:

```
QFATAL : LauncherTest::testLauncher(windll) ASSERT failure in class QString __cdecl GammaRay::WinDllInjector::fixProbeDllPath(const class QString &): "C:\Qt\src\GammaRay\build-dbg\plugins\gammaray\2.9\qt5_9-MSVC-140-x86_64d\gammaray_winloader_wind.dll does not exists!", file ..\launcher\core\injector\windllinjector.cpp, line 208
```

The lib is called: `plugins\gammaray\2.9\qt5_9-MSVC-140-x86_64d\gammaray_winloaderd.dll`

launcher/core/injector/windllinjector.cpp
```diff
@@ -204,7 +204,8 @@ void WinDllInjector::processExited(int exitCode)
 QString WinDllInjector::fixProbeDllPath(const QString &probeDll)
 {
     QString dllPath = QDir::toNativeSeparators(probeDll);
-    dllPath.replace(QLatin1String("gammaray_probe"), QLatin1String("gammaray_probe_win"));
```
milianw: if you replace `gammaray_probe` with `gammaray_probe_win` it works for me

TheOneRing: why do we allow two different plugin layouts ....

milianw: I just realize - this makes the whole replace a no-op, so the whole replace line can be removed?! I still don't know what you mean by different plugin layouts.

milianw: nvm the above comment, it isn't a noop:

```
dllPath.replace(QLatin1String("gammaray_probe_win"), QLatin1String("gammaray_winloader"));
```

milianw: With that fix in, both the release build and the debug build work for me. Most notably, the debug probe build as part of the release build also works.

milianw:
still does not work:

```
QFATAL : LauncherTest::testLauncher(windll) ASSERT failure in class QString __cdecl GammaRay::WinDllInjector::fixProbeDllPath(const class QString &): "C:\Qt\src\GammaRay\build-dbg\plugins\gammaray\2.9\qt5_9-MSVC-140-x86_64d\gammaray_winloader_wind.dll does not exists!", file ..\launcher\core\injector\windllinjector.cpp, line 209
FAIL!  : LauncherTest::testLauncher(windll) Received a fatal error.
```

The `_win` suffix is still wrong.

milianw:
To make this clear, here's an excerpt from my `config-gammaray.h`:

```
#define GAMMARAY_PROBE_BASENAME "gammaray_probe"
```

milianw:
OK, the issue was an outdated `gammaray_probe_wind.dll` in my build dir, which was loaded via the wildcard pattern... Removing that one fixes this issue for me and the patch here works.

PR #482, opened 2018-01-05
Wip/remoteview on software renderer

akreuzkamp:


vkrause:
This seems to work in general, but I'm seeing some rendering issues when toggling target decoration in the client when using the sw renderer. Disabling target decorations results in some garbage in undamaged areas, enabling it requires changing the item selection to take effect.

PR #486, opened 2018-01-26
Remove many (not all) of the automagic dependencies

Petross404:
It would be better for the developers, the packagers and finally the users, if CMakeLists.txt din't rely on whether a package is installed to use it, but rather the user/developer should have fine control on what gammaray will look up as a dependency.

For more, there is a relevant article on Gentoo [wiki](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies). Thank you.


krf:
This is *at most* better for packagers; anyone else will be annoyed having to enable all individual options him or herself.

Just a quick heads up as I'm almost half-asleep now: I don't think this change should go in; it merely adds boilerplate code.

You can disable detection of individual (CMake) packages in a standard way via `-DCMAKE_DISABLE_FIND_PACKAGE` if needed (cf. https://blogs.kde.org/2011/11/11/cool-new-stuff-cmake-286-3-standard-way-disable-optional-packages)

Petross404:
Yes, I don't like the code either. I am already informed by the gentoo developers about the cmake option you linked, thank you.

The package already  works, I make use of the `DCMAKE_DISABLE_FIND_PACKAGE_MyPackageName=TRUE` option and there is no problem besides the ebuild's size. I will contact them, if they find it a good option we can provide our patch for cmakelists.txt. 

Thank you for your time.

krf:
I guess we can close this then? Doing so. Thanks!

PR #491, opened 2018-03-05
Restore compatibility with CMake 3.0

akreuzkamp:
Might be obsolete already, but 2.9 doesn't mention cmake 3.1 as requirement.

### DIFF
--- a/plugins/qmlsupport/CMakeLists.txt
+++ b/plugins/qmlsupport/CMakeLists.txt
@@ -11,15 +11,15 @@ set(gammaray_qmlsupport_srcs
   qmltypeextension.cpp
 )
 
+if(NOT Qt5Qml_VERSION VERSION_LESS 5.10)
+    set(gammaray_qmlsupport_srcs ${gammaray_qmlsupport_srcs} qmlbindingprovider.cpp)
+endif()
+
 gammaray_add_plugin(gammaray_qmlsupport JSON gammaray_qmlsupport.json SOURCES ${gammaray_qmlsupport_srcs})
 target_include_directories(gammaray_qmlsupport SYSTEM PRIVATE ${Qt5Qml_PRIVATE_INCLUDE_DIRS})
 target_link_libraries(gammaray_qmlsupport gammaray_core Qt5::Qml)
 endif()
 
-
-if(NOT Qt5Qml_VERSION VERSION_LESS 5.10)
-    target_sources(gammaray_qmlsupport PUBLIC qmlbindingprovider.cpp)
-endif()
 
 # ui plugin
 if(GAMMARAY_BUILD_UI)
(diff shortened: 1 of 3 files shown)


plugins/qmlsupport/CMakeLists.txt
```diff
@@ -11,16 +11,16 @@ set(gammaray_qmlsupport_srcs
   qmltypeextension.cpp
 )
 
+if(NOT Qt5Qml_VERSION VERSION_LESS 5.10)
+    set(gammaray_qmlsupport_srcs ${gammaray_qmlsupport_srcs} qmlbindingprovider.cpp)
```
ferdnyc, line 15: Alternative:
```suggestion
    list(APPEND gammaray_qmlsupport_srcs qmlbindingprovider.cpp)
```
...I don't really have any insights into whether there's a performance difference, or which form would come out ahead if there is. I just know that `list(APPEND...` is available in any CMake 3.0+ (at least), and it avoids one repetition of the variable name.

plugins/quickinspector/CMakeLists.txt
```diff
@@ -28,15 +28,15 @@ if(Qt5Quick_FOUND)
     textureextension/qsgtexturegrabber.cpp
   )
 
+  if(NOT Qt5Quick_VERSION VERSION_LESS 5.7)
+    set(gammaray_quickinspector_srcs ${gammaray_quickinspector_srcs} quickimplicitbindingdependencyprovider.cpp)
```
ferdnyc, line 32:
```suggestion
    list(APPEND gammaray_quickinspector_srcs quickimplicitbindingdependencyprovider.cpp)
```

tests/CMakeLists.txt
```diff
@@ -425,19 +425,18 @@ if(NOT GAMMARAY_CLIENT_ONLY_BUILD AND Qt5Core_FOUND AND NOT Qt5Core_VERSION_MINO
...
-    )
-    target_include_directories(bindinginspectortest SYSTEM PRIVATE ${Qt5Quick_PRIVATE_INCLUDE_DIRS})
-    target_link_libraries(bindinginspectortest gammaray_core Qt5::Quick)
-    target_sources(bindinginspectortest PUBLIC
       ${CMAKE_SOURCE_DIR}/plugins/quickinspector/quickimplicitbindingdependencyprovider.cpp
     )
     if(NOT Qt5Qml_VERSION VERSION_LESS 5.10)
-      target_sources(bindinginspectortest PUBLIC
+      set(bindinginspectortest_srcs ${bindinginspectortest_srcs}
         ${CMAKE_SOURCE_DIR}/plugins/qmlsupport/qmlbindingprovider.cpp)
```
ferdnyc, line 435:
```suggestion
      list(APPEND bindinginspectortest_srcs
        ${CMAKE_SOURCE_DIR}/plugins/qmlsupport/qmlbindingprovider.cpp)
```

ferdnyc (commented):
An alternative to re-`set()` ing the source list is to use `list(APPEND ...)` to tack on additional sources.

PR #503, opened 2018-09-18
Property filtering

akreuzkamp:
Adds the filtering infrastructure, Filters for anchors and a anchors-property adaptor to add the anchors the way, QtQuick wants us to.

This is the first step to implement the new censorship infrastructure required by the new EU copyright regulation.


akreuzkamp:
don't merge, I introduced a showstopper-bug. Will need to fix it first.

PR #505, opened 2018-09-19
Adapt example and docs to the improved batcher

### LINKED ISSUE #498: Qt Quick 2 batching example / documentation are outdated and misleading
Starting with Qt 5.8 the Rectangle item in Qt Quick uses per-vertex colors. Also a smarter strategy of batching is implemented by the engine. This results in the example being drawn in a few draw calls (instead of one per rectangle), so not actually showing the problem it's supposed to show; the documentation also warning about using Rotation seems to be wrong, as that doesn't affect anything.

akreuzkamp:
Since Qt 5.8 the batcher is able to batch rectangles that differ in
color, have rotations applied or have opacity <1 even if overlapping
(as long as all have the same opacity).
So with this patch the docs don't spread fake news about batching any
longer. Also the example was adapted to use different opacity values for
all the rectangles to trigger the unmerged batch performance issue.

Fixes #498

### DIFF
--- a/docs/manual/examples/quick-batching.qdoc
+++ b/docs/manual/examples/quick-batching.qdoc
@@ -25,7 +25,9 @@
 
     \section1 Problem
 
-    The example application shows two custom Qt Quick sliders consisting of flat colored rectangles.
+    The example application shows two custom Qt Quick sliders consisting of flat colored rectangles
+    forming a gradient from green to blue. Inactive rectangles (that is rectangles above the handle)
+    are drawn smaller and slightly greyed out.
 
     \snippet quick-batching/quick-batching.qml Slider setup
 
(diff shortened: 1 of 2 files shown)


docs/manual/examples/quick-batching.qdoc
```diff
@@ -62,40 +62,42 @@
...
-    In order to identify the culprit, we can move up the hierarchy to the first node introducing this
-    transformation.
+    Comparing the subtrees belonging to active and inactive rectangles scenegraph-node tree view,
+    we notice that the inactive ones have an Opacity node, while the active ones don't. The opacity
+    node indicates transparency.
 
-    In order to match this to the actual code causing the problem, switch back to the item tree view,
-    and use the context menu action "Go to declaration", which will open a code editor around the
-    following piece of code:
+    now we can check the opacity property in the \l {Properties} tab of the Item view, so we switch back to
```
dangelog: Capital N?

docs/manual/examples/quick-batching.qdoc
```diff
@@ -62,40 +62,42 @@
...
+    rectangles.
 
-    Just removing the transformations is of course not a proper fix (as it changes the visual appearance
-    of the application), it merely verifies that it's these rotations causing our performance problem.
-    You would now need to refactor the code in a way that it achieves the same behavior and visual
-    appearance without the use of rotational transformations.
+    Just removing the opacity is of course not a proper fix, it merely verifies that it's these opacities
+    causing our performance problem.
+    To retain the same visual appearance as with opacity set, you could e.g. use the alpha component of
+    the color property to obtain the transparency, without impairing the batching.
```
dangelog: Does this actually work, in the given example? Does it exploit the fact that the items don't overlap, so they can be batched in the alpha pass?

akreuzkamp: Yes it did work (previously I wrote, it wouldn't until I realized it does actually) but no, it doesn't exploit the fact that the items don't overlap. I explicitly made them overlap with each rectangle having a unique alpha-value, still it was able to batch them (or at least the batch-visualizer showed them as batched)

docs/manual/examples/quick-batching.qdoc
akreuzkamp: yes.

PR #514, opened 2018-11-14
Problem reporter

akreuzkamp:
This PR adds a global focal point for tools to report issues that they (already) find during their analysis, with the possibility for the user to actively trigger scans through all objects for certain kinds of issues.

It contains a tool with a UI for listing the issues, filtering the checkers and triggering the scans and includes scans for
* Property Bindings (implemented in the Object Inspector and fed by data from the qmlsupport and quickinspector plugins)
* Shortcut duplicates (provided by the Action Inspector)
* common issues in QMetaType treatment (provided by the MetaObjectBrowser)

Still to do is Direct-Crossthread-Connections, but that is pretty independent of the work already done.

### DIFF
--- a/ui/tools/problemreporter/problemreporterwidget.cpp
+++ b/ui/tools/problemreporter/problemreporterwidget.cpp
@@ -197,7 +197,10 @@ void ProblemReporterWidget::problemViewContextMenu(const QPoint &p)
 
     QMenu menu;
     ContextMenuExtension ext(objectId);
-    ext.setLocation(ContextMenuExtension::GoTo, index.data(ProblemModelRoles::SourceLocationRole).value<SourceLocation>());
+    auto sourceLocations = index.data(ProblemModelRoles::SourceLocationRole).value<QVector<SourceLocation>>();
+    foreach (const SourceLocation &sourceLocation, sourceLocations) {
+        ext.setLocation(ContextMenuExtension::GoTo, sourceLocation);
+    }
     ext.populateMenu(&menu);
 
     menu.exec(ui->problemView->viewport()->mapToGlobal(p));
--- a/ui/tools/problemreporter/problemreporterwidget.cpp
+++ b/ui/tools/problemreporter/problemreporterwidget.cpp
@@ -74,23 +74,24 @@ protected:
         const int hMargin = viewItemOption->widget->style()->pixelMetric(QStyle::PM_FocusFrameHMargin)
                           + viewItemOption->widget->style()->pixelMetric(QStyle::PM_CheckBoxLabelSpacing);
         const QRect checkboxRect = style->subElementRect(QStyle::SE_CheckBoxIndicator, &checkboxOption, viewItemOption->widget);
-        const QRect rect = option.rect.adjusted(checkboxRect.width() + hMargin, vMargin, -hMargin, -vMargin);
-        const int lineHeight = rect.height() / 2;
-        const QRect titleRect(rect.left() + hMargin,
-                              rect.top() + vMargin,
-                              rect.width(),
-                              lineHeight);
-        const QRect authorRect(rect.left() + hMargin,
-                               rect.top() + vMargin + lineHeight,
-                               rect.width(),
-                               lineHeight);
+        const QRect textRect = option.rect.adjusted(checkboxRect.width() + hMargin, vMargin, -hMargin, -vMargin);
+        const QFontMetrics metrics(option.font);
+        const QRect titleRect(textRect.left(),
+                              textRect.top(),
+                              textRect.width(),
+                              metrics.height());
+        const QRect descriptionRect = metrics.boundingRect(textRect, Qt::TextWordWrap, description)
+                                             .translated(0, vMargin + titleRect.height());
+
+        const QRect descriptionRect2 = metrics.boundingRect(viewItemOption->rect.adjusted(checkboxRect.width(),0,0,0), Qt::TextWordWrap, description);
 
         painter->save();
         painter->drawText(titleRect, Qt::AlignLeft | Qt::AlignVCenter, title);
         painter->restore();
         painter->setOpacity(0.5);
-        painter->drawText(authorRect, Qt::AlignLeft | Qt::AlignVCenter, description);
+        painter->drawText(descriptionRect, Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap, description);
         painter->restore();
+
 
         style->drawControl(QStyle::CE_CheckBox, &checkboxOption, painter, viewItemOption->widget);
     }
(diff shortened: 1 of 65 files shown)


ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <QStyleOptionViewItemV4>
+
+
+#if QT_VERSION < QT_VERSION_CHECK(5, 5, 0)
+Q_DECLARE_METATYPE(Qt::CheckState)
+#endif
+
+using namespace GammaRay;
+
+class ProblemFilterDelegate : public QStyledItemDelegate
```
akreuzkamp: this class feels like the worst code I've written this year. So, please check it extra carefully and ideally tell me, how to get the rendering right. :)

akreuzkamp:
the failing builds seem to be a matter of doing a clean build. I'll trigger one, when the review is done

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+                              textRect.width(),
+                              metrics.height());
+        const QRect descriptionRect = metrics.boundingRect(textRect, Qt::TextWordWrap, description)
+                                             .translated(0, vMargin + titleRect.height());
+
+        const QRect descriptionRect2 = metrics.boundingRect(viewItemOption->rect.adjusted(checkboxRect.width(),0,0,0), Qt::TextWordWrap, description);
+
+        painter->save();
+        painter->drawText(titleRect, Qt::AlignLeft | Qt::AlignVCenter, title);
+        painter->restore();
```
dfaure-kdab:
```suggestion
```

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+                              textRect.top(),
+                              textRect.width(),
+                              metrics.height());
+        const QRect descriptionRect = metrics.boundingRect(textRect, Qt::TextWordWrap, description)
+                                             .translated(0, vMargin + titleRect.height());
+
+        const QRect descriptionRect2 = metrics.boundingRect(viewItemOption->rect.adjusted(checkboxRect.width(),0,0,0), Qt::TextWordWrap, description);
+
+        painter->save();
+        painter->drawText(titleRect, Qt::AlignLeft | Qt::AlignVCenter, title);
```
dfaure-kdab: vertical centering in a rectangle that's exactly as high as the text? That seems like a no-op...

IIUC you could use AlignTop and use textRect instead of titleRect here.

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+        const QRect textRect = option.rect.adjusted(checkboxRect.width() + hMargin, vMargin, -hMargin, -vMargin);
+        const QFontMetrics metrics(option.font);
+        const QRect titleRect(textRect.left(),
+                              textRect.top(),
+                              textRect.width(),
+                              metrics.height());
+        const QRect descriptionRect = metrics.boundingRect(textRect, Qt::TextWordWrap, description)
+                                             .translated(0, vMargin + titleRect.height());
+
+        const QRect descriptionRect2 = metrics.boundingRect(viewItemOption->rect.adjusted(checkboxRect.width(),0,0,0), Qt::TextWordWrap, description);
```
dfaure-kdab:
```suggestion
```

unused

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+        painter->save();
```
dfaure-kdab: This isn't incorrect, but it would be faster to just save/restore the opacity in a local variable, rather than the full painter state.

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+        const int hMargin = viewItemOption->widget->style()->pixelMetric(QStyle::PM_FocusFrameHMargin)
+                          + viewItemOption->widget->style()->pixelMetric(QStyle::PM_CheckBoxLabelSpacing);
+        const QRect checkboxRect = style->subElementRect(QStyle::SE_CheckBoxIndicator, &checkboxOption, viewItemOption->widget);
+        const QRect textRect = option.rect.adjusted(checkboxRect.width() + hMargin, vMargin, -hMargin, -vMargin);
+        const QFontMetrics metrics(option.font);
+        const QRect titleRect(textRect.left(),
+                              textRect.top(),
+                              textRect.width(),
+                              metrics.height());
+        const QRect descriptionRect = metrics.boundingRect(textRect, Qt::TextWordWrap, description)
```
dfaure-kdab: You're wrapping the text in a rectangle that includes the space for the title.
Plus, you're doing that work twice (once here and once during painting).

Why not just compute the available rect, and wrap during drawText?
    const QRect descriptionRect = textRect.adjust(0, metrics.height(), 0, 0);

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+        painter->save();
+        painter->drawText(titleRect, Qt::AlignLeft | Qt::AlignVCenter, title);
+        painter->restore();
+        painter->setOpacity(0.5);
+        painter->drawText(descriptionRect, Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap, description);
+        painter->restore();
+
+
+        style->drawControl(QStyle::CE_CheckBox, &checkboxOption, painter, viewItemOption->widget);
```
dfaure-kdab: isn't there a risk that this overwrites your text drawing? I'd do this first...

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+        const QString title = index.data(Qt::DisplayRole).toString();
+        const QString description = index.data(Qt::ToolTipRole).toString();
+        const QFontMetrics metrics(option.font);
+        const int vMargin = viewItemOption->widget->style()->pixelMetric(QStyle::PM_FocusFrameVMargin);
+        const int hMargin = viewItemOption->widget->style()->pixelMetric(QStyle::PM_FocusFrameHMargin)
+                          + viewItemOption->widget->style()->pixelMetric(QStyle::PM_CheckBoxLabelSpacing);
+        const QRect checkboxRect = viewItemOption->widget->style()->subElementRect(QStyle::SE_CheckBoxIndicator, &checkboxOption, viewItemOption->widget);
+        const QRect titleRect = metrics.boundingRect(title);
+        const int textAvailableWidth = qobject_cast<const QListView*>(viewItemOption->widget)->width() // FIXME this is the wrong size if the view has a horizontal scrollbar...
```
dfaure-kdab: unchecked qobject_cast = either should be static_cast, or should have a null check.

Maybe you want to cast to QAbstractScrollArea more generally, and then call viewport()->width()? Not 100% sure.

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+            - 2 * viewItemOption->widget->style()->pixelMetric(QStyle::PM_FocusFrameHMargin) - checkboxRect.width();
+        const QRect descriptionRect = metrics.boundingRect(QRect(0,0,textAvailableWidth, 0), Qt::TextWordWrap, description);
```
dfaure-kdab: This rect has a height of 0... a bit surprising, but if it works....

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+protected:
+    void paint(QPainter *painter, const QStyleOptionViewItem &option,
+               const QModelIndex &index) const override
+    {
+        const QStyleOptionViewItemV4 *viewItemOption =
+           qstyleoption_cast<const QStyleOptionViewItemV4 *>(&option); // This is for Qt4 compatibility. Should be a noop in Qt5
+        QStyle *style = viewItemOption->widget ? viewItemOption->widget->style() : QApplication::style();
+        QStyleOptionButton checkboxOption = copyStyleOptions(option, index);
+        painter->save();
+        QString title = index.data(Qt::DisplayRole).toString();
```
dfaure-kdab:
```suggestion
        const QString title = index.data(Qt::DisplayRole).toString();
```

ui/tools/problemreporter/problemreporterwidget.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+protected:
+    void paint(QPainter *painter, const QStyleOptionViewItem &option,
+               const QModelIndex &index) const override
+    {
+        const QStyleOptionViewItemV4 *viewItemOption =
+           qstyleoption_cast<const QStyleOptionViewItemV4 *>(&option); // This is for Qt4 compatibility. Should be a noop in Qt5
+        QStyle *style = viewItemOption->widget ? viewItemOption->widget->style() : QApplication::style();
+        QStyleOptionButton checkboxOption = copyStyleOptions(option, index);
+        painter->save();
```
dfaure-kdab:
```suggestion
```
unnecessary, you're not changing the painter state here

PR #518, opened 2018-12-04
Add more debugging options as environment vars

akreuzkamp:
Adds an environment variable GAMMARAY_DEBUG with the values 'GDB',
'GDB_NORUN' and 'RR'. 'GDB' is equivalent to setting GAMMARAY_GDB=1,
'GDB_NORUN' does the same, but does not automatically run the
application and 'RR' calls the application with `rr record`.


dantti:
@akreuzkamp to merge this?

Waqar144:
merging this as already approved.

PR #525, opened 2019-01-09
CMake QNX700 toolchain file fixes

krf:


krf:
Just tested with the "Toolchain-QNX700-armv7.cmake" toolchain file => this seems to build fine.

PR #526, opened 2019-01-14
Fix the Qt modules for Qt 5.12.1

e4z9:
Since Qt 5.12.1 (qtbase e0926ca4260953a70cdb2a6f84da35517ead1349)
Qt expects modules to provide a "QT.<module>.module" entry.
This is used for the actual linker line.


e4z9:
Link to the Qt upstream change on Gerrit: 
https://codereview.qt-project.org/248684

e4z9:
> Mind if I upstream this to KDE's Extra CMake Modules too, where this code actually comes from?

please go ahead!

vkrause:
FTR, submitted upstream as https://phabricator.kde.org/D18243

PR #534, opened 2019-02-22
Fix infinite recursion when app has its own message handler.

dfaure-kdab:
This is because gammaray's message handler is being installed twice.

Testcase: `gammaray zanshin`


dfaure-kdab:
Approved by Volker in chat; pushed directly to 2.10 branch and merged to master.

PR #538, opened 2019-03-02
ObjectInspector: Add thread affinity checker

chgans:
Some new checks i've been thinking about, i think they are sound, but i'm open to ideas on how to improve this.
When run on a empty Qt project (eg, w/ just an app and a window), this gives two warnings:
`The thread QDBusConnectionManager[this=0x7f1a5516ad80] has affinity with itself.`
`The thread QThread[this=0x5560ce4e0680] has affinity with itself.`
So maybe this new check should not be enabled by default. I haven't looked yet where to disable it.

### DIFF
--- a/core/tools/objectinspector/objectinspector.cpp
+++ b/core/tools/objectinspector/objectinspector.cpp
@@ -55,6 +55,7 @@
 #include <QMetaMethod>
 
 #include <QMutexLocker>
+#include <QThread>
 
 using namespace GammaRay;
 
@@ -80,14 +81,17 @@ ObjectInspector::ObjectInspector(Probe *probe, QObject *parent)
             this, &ObjectInspector::objectSelected);
 
     ProblemCollector::registerProblemChecker("com.kdab.GammaRay.ObjectInspector.BindingLoopScan",
-                                          "Binding Loops",
-                                          "Scans all QObjects for binding loops",
-                                          &BindingAggregator::scanForBindingLoops);
+                                             "Binding Loops",
+                                             "Scans all QObjects for binding loops",
+                                             &BindingAggregator::scanForBindingLoops);
     ProblemCollector::registerProblemChecker("com.kdab.GammaRay.ObjectInspector.ConnectionsCheck",
-                                          "Connection issues",
-                                          "Scans all QObjects for direct cross-thread and duplicate connections",
-                                          &ObjectInspector::scanForConnectionIssues);
-
+                                             "Connection issues",
+                                             "Scans all QObjects for direct cross-thread and duplicate connections",
+                                             &ObjectInspector::scanForConnectionIssues);
+    ProblemCollector::registerProblemChecker("com.kdab.GammaRay.ObjectInspector.ThreadAffinityCheck",
+                                             "Threading issues",
+                                             "Scans all QObjects for thread affinity issues",
+                                             &ObjectInspector::scanForThreadAffinityIssues);
 }
 
 void ObjectInspector::objectSelectionChanged(const QItemSelection &selection)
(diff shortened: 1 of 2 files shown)


core/tools/objectinspector/objectinspector.cpp
```diff
@@ -206,3 +210,61 @@ void ObjectInspector::scanForConnectionIssues()
...
+
+    QMutexLocker lock(Probe::objectLock());
+    for (const auto object : objects) {
+        if (!Probe::instance()->isValidObject(object)) {
+            continue;
+        }
+
+        const auto objectName = Util::displayString(object);
+        if (object == object->thread()) {
+            const auto objectName = Util::displayString(object);
```
vkrause: duplicating objectName from two lines above?

core/tools/objectinspector/objectinspector.cpp
```diff
@@ -206,3 +210,61 @@ void ObjectInspector::scanForConnectionIssues()
...
+            Problem problem;
+            problem.severity = Problem::Warning;
+            problem.description = QStringLiteral("The thread %1 has affinity with itself.").arg(objectName);
+            problem.object = ObjectId(object);
+            // p.location = bindingNode->sourceLocation(); //TODO can we get source locations of new-statements?
```
vkrause: See Probe::objectCreationSourceLocation().

core/tools/objectinspector/objectinspector.cpp
```diff
@@ -206,3 +210,61 @@ void ObjectInspector::scanForConnectionIssues()
...
+            problem.problemId = QStringLiteral("com.kdab.GammaRay.ObjectInspector.ThreadAffinityCheck.%1:%2")
+                    .arg(QString::number(reinterpret_cast<quintptr>(object)),
+                         QString::number(reinterpret_cast<quintptr>(parent)));
+            problem.findingCategory = Problem::Scan;
+            ProblemCollector::addProblem(problem);
+        }
+
+        if (parent->inherits("QThread") && object->thread() != object->parent()) {
```
vkrause: qobject_cast<QThread*>(parent) would be an alternative to inherits that doesn't require string comparison

core/tools/objectinspector/objectinspector.cpp
chgans: Thanks for the tip. I just realised that the `Problem` class uses a `QVector<SourceLocation>`, while it would have been nice to use `Execution::Trace`.
Is there any reason for that?

For tis PR I will just use `problem.locations.append(probe->objectCreationSourceLocation(object));`

akreuzkamp: The idea of QVector<SourceLocation> is not to hold a trace, but to hold different relevant locations. For a duplicated connect issue, you'd want to see the locations of both connects, not just the second (I'm not sure, this is actually implemented yet, but that's why locations is a vector).
Using `problem.locations.append(probe->objectCreationSourceLocation(object));` is exactly how it's meant to be used :)

chgans: OK, makes sense, and for the backtrace, there's always the "Show in object tool".

core/tools/objectinspector/objectinspector.cpp
```diff
@@ -206,3 +210,61 @@ void ObjectInspector::scanForConnectionIssues()
         }
     }
 }
+
+void ObjectInspector::scanForThreadAffinityIssues()
+{
+    const auto probe = Probe::instance();
+    const auto &objects = probe->allQObjects();
```
chgans: Quick question: I've used the `scanForConnectionIssues` as a template, and i've just noticed that `probe->allQObjects();` is called just before locking the probe, wouldn't it be safer/simpler, to first lock the probe?

akreuzkamp: As far as I see it, the current form is correct. That's only the case, as long as the signature of allObjects() doesn't get change to return a copy instead of a const-ref, though. So moving the lock up, would actually be more forward-safe. :)

chgans: OK, i'll leave it like this, unless you request the change.

PR #541, opened 2019-03-18
Add custom displayString case for QJsonObject

nicolasfella:
Workaround for QTBUG-73437

### DIFF
--- a/core/varianthandler.cpp
+++ b/core/varianthandler.cpp
@@ -53,6 +53,7 @@
 #include <QVector2D>
 #include <QVector3D>
 #include <QVector4D>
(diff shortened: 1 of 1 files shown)


core/varianthandler.cpp
```diff
@@ -322,6 +323,15 @@ QString VariantHandler::displayString(const QVariant &value)
         return (*it.value())(value);
     }
 
+    if (value.canConvert<QJsonObject>()) {
```
vkrause: This probably shouldn't be canConvert() but userType() == qMetaTypeId<QJsonObject>(), +2 otherwise.

PR #543, opened 2019-03-19
Chgans/allow preload injector override

chgans:
This the PR for https://mail.kdab.com/pipermail/gammaray-interest/2019-March/000317.html.

It is untested so far, and i won't be able to test it before April. Nonetheless, you might want to comment on that.

### DIFF
--- a/launcher/core/injector/preloadinjector.cpp
+++ b/launcher/core/injector/preloadinjector.cpp
@@ -68,7 +68,7 @@ bool PreloadInjector::launch(const QStringList &programAndArgs, const QString &p
     // ASAN requires to be loaded first, so check if the target uses that
     // and if so inject it before GammaRay
     QStringList ldPreload;
-    foreach (const auto &lib, LibraryUtil::dependencies(exePath)) {
+    for (const auto &lib: LibraryUtil::dependencies(exePath)) {
         if (lib.contains("libasan.so") || lib.contains("libclang_rt.asan")) {
             ldPreload.push_back(QString::fromLocal8Bit(lib));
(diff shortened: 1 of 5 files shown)


launcher/core/injector/preloadinjector.cpp
```diff
@@ -51,13 +55,15 @@ bool PreloadInjector::launch(const QStringList &programAndArgs, const QString &p
 {
     Q_UNUSED(probeFunc);
 
+    const QString actualProbeDll = m_probeDllOverride.isEmpty() ? probeDll : m_probeDllOverride;
```
akreuzkamp: This would be better preserved at the caller side of this function, i.e. in Launcher::start. That requires the use of a separated commandline option though.

vkrause:
I think you are right, separating the command line arguments affecting the probe from those affecting the injector makes sense. The approach in general is fine IMHO, as discussed by email already.

chgans:
Oops, forgot about this PR....
Came back here as i wanted to check if Android is supported, then look at what's cooking and found my own un-merged PR :)

chgans:
I have signed NDA ages ago, and I have already contributed to GammaRay

chgans:
When i follow the link, i get:
```

Error

There is no CLA to sign for KDAB/GammaRay

(could not find linked item for owner KDAB and repo GammaRay)

```

winterz:
yes, I'm sorry.  I'm trying to use cla-assistant webhook to make things easier than "paper".  still learning how to use it and I need to pre-approved those who already signed.

PR #547, opened 2019-04-29
Allow to build GammaRay when Qt5OpenGL is not present

### LINKED ISSUE #546: plugin quickinspector not building when QT library is build with no-opengl support QT_NO_OPENGL
While using Yocto to build for a x86 environment, Qt was build with --no-opengl option . Which means QT_NO_OPENGL will be defined QT global include. The quickinspector plugin uses the GL attributes in various files without checking for this. A similar fix was done in guisupport plugin 2 years back.

gmarull:
The changes in this PR allow to build GammaRay when Qt5 is built without OpenGL support.

Fixes #546 

NOTE: Due to my little knowledge about the GammaRay codebase, some components may have been unnecessarily removed from the non-OpenGL build.

### DIFF
--- a/3rdparty/kuserfeedback/core/openglinfosource.cpp
+++ b/3rdparty/kuserfeedback/core/openglinfosource.cpp
@@ -19,7 +19,7 @@
 #include "openglinfosource_p.h"
 
 #include <QVariant>
-#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0) && !defined(QT_NO_OPENGL)
 #include <QOpenGLContext>
 #include <QOpenGLFunctions>
 #include <QSurfaceFormat>
@@ -42,7 +42,7 @@ QVariant OpenGLInfoSource::data()
 {
     QVariantMap m;
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0) && !defined(QT_NO_OPENGL)
     QOpenGLContext context;
     if (context.create()) {
         QWindow window;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -391,6 +391,7 @@ find_package(Qt5 NO_MODULE QUIET OPTIONAL_COMPONENTS
   WebEngineWidgets
   Widgets
   WaylandCompositor
+  OpenGL
 )
 # Find these 'exotic' Qt modules without using find_package(... COMPONENTS)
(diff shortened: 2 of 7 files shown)


3rdparty/kuserfeedback/CMakeLists.txt
```diff
@@ -24,6 +23,11 @@ list(APPEND gammaray_userfeedback_srcs
     core/auditloguicontroller.cpp
     widgets/auditlogbrowserdialog.cpp
 )
+if (Qt5OpenGL_FOUND)
+    list(APPEND gammaray_userfeedback_srcs
+        core/openglinfosource.cpp
+    )
+endif()
```
akreuzkamp: Don't remove openglinfosource.cpp completely. The cpp file already contains an #if to check for Qt >= 5.0.0, reuse that to also check for `!defined(QT_NO_OPENGL)`

gmarull: @akreuzkamp Thanks for the feedback. Just fixed the code ;)!

ui/mainwindow.cpp
```diff
@@ -52,7 +52,9 @@
 #include <3rdparty/kuserfeedback/widgets/notificationpopup.h>
 #include <3rdparty/kuserfeedback/core/applicationversionsource.h>
 #include <3rdparty/kuserfeedback/core/compilerinfosource.h>
+#ifndef QT_NO_OPENGL
```
akreuzkamp: remove, we can now always use openglinfosource again.

ui/mainwindow.cpp
```diff
@@ -352,7 +354,9 @@ void MainWindow::setupFeedbackProvider()
     m_feedbackProvider->addDataSource(new KUserFeedback::QtVersionSource);
     m_feedbackProvider->addDataSource(new KUserFeedback::StartCountSource);
     m_feedbackProvider->addDataSource(new KUserFeedback::UsageTimeSource);
+#ifndef QT_NO_OPENGL
```
akreuzkamp: remove

ui/mainwindow.cpp
gmarull: Fixed

ui/mainwindow.cpp
gmarull: Fixed

vkrause:
Thanks! Integrated in 2.11 with a few variations:
- fixed 3rdparty code upstream and resynced with that
- changed the CMake side to not introduce a dependency on Qt5::OpenGL, but rather check if Qt was build with QT_NO_OPENGL (the latter implies the former).
- less aggressive removal of client-side code not depending on OpenGL

gmarull:
Thanks!

PR #548, opened 2019-04-30
Add Event Monitor Plugin

luminosuslight:
This PR adds a new plugin that shows all Qt events in the target application.

### DIFF
--- a/core/metaobjectrepository.cpp
+++ b/core/metaobjectrepository.cpp
@@ -37,6 +37,7 @@
 #include <QEasingCurve>
 #include <QFile>
 #include <QObject>
+#include <private/qobject_p.h>
 #include <QSortFilterProxyModel>
 #include <QStringList>
 #include <QThread>
@@ -184,6 +185,9 @@ void MetaObjectRepository::initIOTypes()
     MO_ADD_METAOBJECT1(QSaveFile, QFileDevice);
 }
 
+Q_DECLARE_METATYPE(const QObject*)
+Q_DECLARE_METATYPE(void**)
+
 void MetaObjectRepository::initQEventTypes()
 {
     MetaObject *mo = nullptr;
@@ -206,6 +210,12 @@ void MetaObjectRepository::initQEventTypes()
 
     MO_ADD_METAOBJECT1(QDeferredDeleteEvent, QEvent);
     MO_ADD_PROPERTY_RO(QDeferredDeleteEvent, loopLevel);
+
+    MO_ADD_METAOBJECT1(QMetaCallEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, id);
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, sender);
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, signalId);
+//    MO_ADD_PROPERTY_RO(QMetaCallEvent, args);  // unusable atm because type is void**
 }
 
 MetaObjectRepository *MetaObjectRepository::instance()
--- a/core/metaobjectrepository.cpp
+++ b/core/metaobjectrepository.cpp
@@ -67,6 +67,7 @@ void MetaObjectRepository::initBuiltInTypes()
     m_initialized = true;
     initQObjectTypes();
     initIOTypes();
+    initQEventTypes();
 }
 
 Q_DECLARE_METATYPE(QThread::Priority)
@@ -181,6 +182,30 @@ void MetaObjectRepository::initIOTypes()
     MO_ADD_PROPERTY_RO(QFile, symLinkTarget);
 
     MO_ADD_METAOBJECT1(QSaveFile, QFileDevice);
+}
+
+void MetaObjectRepository::initQEventTypes()
+{
+    MetaObject *mo = nullptr;
+    MO_ADD_METAOBJECT0(QEvent);
+    MO_ADD_PROPERTY_RO(QEvent, isAccepted);
+    MO_ADD_PROPERTY_RO(QEvent, spontaneous);
+    MO_ADD_PROPERTY_RO(QEvent, type);
+
+    MO_ADD_METAOBJECT1(QTimerEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QTimerEvent, timerId);
+
+    MO_ADD_METAOBJECT1(QChildEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QChildEvent, child);
+    MO_ADD_PROPERTY_RO(QChildEvent, added);
+    MO_ADD_PROPERTY_RO(QChildEvent, polished);
+    MO_ADD_PROPERTY_RO(QChildEvent, removed);
+
+    MO_ADD_METAOBJECT1(QDynamicPropertyChangeEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QDynamicPropertyChangeEvent, propertyName);
+
+    MO_ADD_METAOBJECT1(QDeferredDeleteEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QDeferredDeleteEvent, loopLevel);
 }
 
 MetaObjectRepository *MetaObjectRepository::instance()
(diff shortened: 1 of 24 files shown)


core/metaobjectrepository.cpp
```diff
@@ -183,6 +185,39 @@ void MetaObjectRepository::initIOTypes()
...
+    MO_ADD_PROPERTY_RO(QDynamicPropertyChangeEvent, propertyName);
+
+    MO_ADD_METAOBJECT1(QDeferredDeleteEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QDeferredDeleteEvent, loopLevel);
+
+    MO_ADD_METAOBJECT1(QMetaCallEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, id);
+//    MO_ADD_PROPERTY_RO(QMetaCallEvent, sender);  // problematic because type is const QObject*
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, signalId);
+//    MO_ADD_PROPERTY_RO(QMetaCallEvent, args);  // unusable atm because type is void**
```
vkrause: Right, and this can never work as-is, this will need custom decoding code for args. So we can remove that here, as well as the metatype declaration for void**.

core/metaobjectrepository.cpp
```diff
@@ -183,6 +185,39 @@ void MetaObjectRepository::initIOTypes()
...
+
+    MO_ADD_METAOBJECT1(QDynamicPropertyChangeEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QDynamicPropertyChangeEvent, propertyName);
+
+    MO_ADD_METAOBJECT1(QDeferredDeleteEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QDeferredDeleteEvent, loopLevel);
+
+    MO_ADD_METAOBJECT1(QMetaCallEvent, QEvent);
+    MO_ADD_PROPERTY_RO(QMetaCallEvent, id);
+//    MO_ADD_PROPERTY_RO(QMetaCallEvent, sender);  // problematic because type is const QObject*
```
vkrause: There is precedent for this, see "const QStyle* QStyle::proxy() const" as supported by the widget plugin. This property is also shown correctly in the property model (and recursively expanded).

plugins/eventmonitor/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# probe part
+if (NOT GAMMARAY_CLIENT_ONLY_BUILD)
+set(gammaray_eventmonitor_plugin_srcs
+  eventmonitor.cpp
+  eventmodel.cpp
+)
+
+gammaray_add_plugin(gammaray_eventmonitor_plugin
+  DESKTOP gammaray_eventmonitor.desktop.in
```
vkrause: anything referring to desktop files is no longer needed, this was for the old Qt4 plugin loader

plugins/eventmonitor/eventmodel.cpp
```diff
@@ -0,0 +1,169 @@
+/*
+  eventmodel.cpp
+
+  This file is part of GammaRay, the Qt application inspection and
+  manipulation tool.
+
+  Copyright (C) 2010-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Milian Wolff <milian.wolff@kdab.com>
```
vkrause: copy/paste mistake (repeats) :)

plugins/eventmonitor/eventmodel.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if (parent.internalId() == TopLevelId && parent.column() == 0)
+        // TODO: add event propagation history as children
+        return 0;
+
+    return 0;
+}
+
+QVariant EventModel::data(const QModelIndex &index, int role) const
+{
+    if (!index.isValid() || index.row() > rowCount() || index.column() > columnCount())
```
vkrause: shouldn't this be ">=" for the row and column index checks?

plugins/eventmonitor/eventmodel.cpp
```diff
@@ -0,0 +1,169 @@
...
+                if (Probe::instance()->isValidObject(event.receiver)) {
+                    return Util::displayString(event.receiver);
+                }
+                return Util::addressToString(event.receiver);
+            }
+            }
+        } else {
+            // TODO: add event propagation history as children
+        }
+    } else if (role == EventModelRole::Sort) {
```
vkrause: The Sort role seems unused?

plugins/eventmonitor/eventmonitor.cpp
```diff
@@ -0,0 +1,187 @@
...
+    switch (type) {
+    case QEvent::NonClientAreaMouseMove:
+    case QEvent::NonClientAreaMouseButtonPress:
+    case QEvent::NonClientAreaMouseButtonRelease:
+    case QEvent::NonClientAreaMouseButtonDblClick:
+    case QEvent::MouseButtonDblClick:
+    case QEvent::MouseButtonPress:
+    case QEvent::MouseButtonRelease:
+    case QEvent::MouseMove:
+        return "QMouseEvent";
```
vkrause: This would benefit from QStringLiteral I think, as this is potentially a hot path (called for every event) and that would avoid memory allocation and text codec conversions here.

plugins/eventmonitor/eventmonitor.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+    if (selection.isEmpty())
+        return;
+    const QModelIndex index = selection.first().topLeft();
+    QVariant eventAttributes = QVariant(index.data(EventModelRole::AttributesRole).value<QVariantMap>());
+
+    m_eventPropertyModel->setObject(eventAttributes);
+}
+
+EventMonitor::~EventMonitor() = default;
```
vkrause: This probably should probably reset s_model to nullptr and unregister the callback. This doesn't matter on application exit, but you can detach the probe from a running application and continue without GammaRay (as well as re-attach later to such a process), for that cleaning up especially also "static" state is important.

plugins/guisupport/guisupport.cpp
```diff
@@ -82,6 +84,14 @@ Q_DECLARE_METATYPE(QPixelFormat::ByteOrder)
 Q_DECLARE_METATYPE(QPixelFormat::ColorModel)
 Q_DECLARE_METATYPE(QPixelFormat::TypeInterpretation)
 Q_DECLARE_METATYPE(QPixelFormat::YUVLayout)
+Q_DECLARE_METATYPE(Qt::MouseButton)
+Q_DECLARE_METATYPE(QFlags<Qt::MouseEventFlag>)
+Q_DECLARE_METATYPE(QTouchEvent::TouchPoint)
+Q_DECLARE_METATYPE(QList<QTouchEvent::TouchPoint>)
+Q_DECLARE_METATYPE(Qt::TouchPointState)
+Q_DECLARE_METATYPE(QFlags<QTouchEvent::TouchPoint::InfoFlag>)
+Q_DECLARE_METATYPE(QFlags<QTouchDevice::CapabilityFlag>)
```
vkrause: This one is declared as Q_FLAG, which should auto-declare a metatype. So that shouldn't be needed.

plugins/guisupport/guisupport.cpp
```diff
@@ -82,6 +84,14 @@ Q_DECLARE_METATYPE(QPixelFormat::ByteOrder)
 Q_DECLARE_METATYPE(QPixelFormat::ColorModel)
 Q_DECLARE_METATYPE(QPixelFormat::TypeInterpretation)
 Q_DECLARE_METATYPE(QPixelFormat::YUVLayout)
+Q_DECLARE_METATYPE(Qt::MouseButton)
+Q_DECLARE_METATYPE(QFlags<Qt::MouseEventFlag>)
```
vkrause: Elements in the Qt namespace typically have metatypes already via a Q_ENUM/Q_FLAG special-case. Sure we need those here?

plugins/guisupport/guisupport.cpp
luminosuslight: You're right, but somehow it doesn't compile without it (`qt_metatype_id’ is not a member of ‘QMetaTypeId<QFlags<QTouchDevice::CapabilityFlag> >`)

plugins/guisupport/guisupport.cpp
luminosuslight: This doesn't compile, either, and even with those lines the MouseEventFlags are not correctly shown in the UI :thinking:

PR #549, opened 2019-05-02
Allow to build without Qt5Svg

gmarull:


akreuzkamp:
Hey, thanks for contributing this! Apparently @vkrause fixed the same thing meanwhile, so I have to reject this, even though the fix is correct. Consider it merged. :)

PR #571, opened 2019-10-03
Add IFW package creation

amantia:
CMake module to create IFW packages


amantia:
Not tested on windows, where cmake says make DESTDIR=foo install might not work as expected, so we might need a special case for windows.
Also it depends on QtInstallPaths.cmake from GammaRay.

winterz:
+1 from me.

PR #579, opened 2019-11-11
Fixing problem with backtrace and uclibc

angeloc:
When compiling on uclibc, backtrace is not always available, so gammaray doesn't compile.

This series trying to fixing it.

### DIFF
--- a/core/execution.cpp
+++ b/core/execution.cpp
@@ -31,7 +31,7 @@
 
 #include <QtGlobal>
 
-#if defined(Q_OS_LINUX) && !defined(Q_OS_ANDROID)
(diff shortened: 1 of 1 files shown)


winterz:
these fixes would be better addressed by the backward-cpp project
https://github.com/bombela/backward-cpp

I will see about making sure our 3rdparty copy is the latest version.

winterz:
maybe something can be done without modifying 3rdparty code. I'll look into it.

angeloc:
@winterz : I updated the pullrequest, the patch regarding backward-cpp is not really needed

PR #582, opened 2019-11-25
Add 'conan' package recipe

renatofilho:


conan/ECM/conanfile.py
```diff
@@ -0,0 +1,53 @@
+###############################################################################
+##  This file is part of GammaRay, the Qt application inspection and
+##  manipulation tool.
+##
+##  Copyright (C) 2015-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+##  Author: Volker Krause <volker.krause@kdab.com>
```
akreuzkamp: This should be your name, no?
(same for the other files)

PR #584, opened 2019-12-04
Fix some GCC 9 warnings

marc-kdab:
- pessimizing std::move
- memcpy to object of non-trivial type

The code in widget3dmodel.h is almost certainly wrong, since it memcpy's
8 bytes into a buffer of 16 bytes, but I left it as-is, because I don't
know if and where it's converted back.

### DIFF
--- a/plugins/widgetinspector/widget3dmodel.h
+++ b/plugins/widgetinspector/widget3dmodel.h
@@ -65,7 +65,8 @@ public:
     inline QString id() const {
         QString str(8, QLatin1Char('0'));
         const quint64 ptr = reinterpret_cast<quint64>(mQWidget.data());
-        std::memcpy(str.data(), &ptr, 8);
+        // this is almost certainly wrong, but keep it for backwards compat:
+        std::memcpy(static_cast<void*>(str.data()), &ptr, 8);
         return str;
     };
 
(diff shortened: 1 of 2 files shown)


plugins/widgetinspector/widget3dmodel.h
```diff
@@ -63,10 +63,7 @@ class Widget3DWidget : public QObject
     // QML does not handle 64bit integers, so use string instead, we only need
     // the value for comparison, we can convert back to quintptr in C++
     inline QString id() const {
-        QString str(8, QLatin1Char('0'));
-        const quint64 ptr = reinterpret_cast<quint64>(mQWidget.data());
-        std::memcpy(str.data(), &ptr, 8);
-        return str;
+        return QString::asprintf("%p", static_cast<const void*>(mQWidget.data()));
```
akreuzkamp: This code isn't equivalent to the old version of the function. This converts the pointer into a user-readable string, while the old version (mis)used the string as a bytearray containing the pointer. It doesn't really matter, though, as it's only used for comparison as far as I see.

dantti:
> The code in widget3dmodel.h is almost certainly wrong, since it memcpy's
> 8 bytes into a buffer of 16 bytes, but I left it as-is, because I don't
> know if and where it's converted back.

@marc-kdab you said you left as-is before you pushed the commit? The patch looks fine to me as the commend above id mentions it needs to be text just because of QML, looking at it's usage as @akreuzkamp said seems fine. Can we merge it?

Waqar144:
merged the first commit locally. Closing.

Issue #586, opened 2019-12-07
Crash when accessing Qt Quick Scene


mitchcurtis:
<img width="532" alt="Screen Shot 2019-12-07 at 2 32 01 pm" src="https://user-images.githubusercontent.com/6002340/70375479-7dd39700-18fe-11ea-8879-b0ccc0145f50.png">

```
Process:               tshnm-iso [99820]
Path:                  /Users/USER/*/tshnm-iso.app/Contents/MacOS/tshnm-iso
Identifier:            org.example.tshnm-iso
Version:               1.0 (1.0.0)
Code Type:             X86-64 (Native)
Parent Process:        Qt Creator [79114]
Responsible:           tshnm-iso [99820]
User ID:               501

Date/Time:             2019-12-07 14:31:08.519 +0100
OS Version:            Mac OS X 10.14.6 (18G1012)
Report Version:        12
Bridge OS Version:     3.0 (14Y906)

Time Awake Since Boot: 520000 seconds
Time Since Wake:       3200 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGABRT)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

External Modification Warnings:
Debugger attached to process.

VM Regions Near 0:
--> 
    __TEXT                 0000000102ff3000-0000000103c0a000 [ 12.1M] r-x/rwx SM=COW  /Users/USER/*/tshnm-iso.app/Contents/MacOS/tshnm-iso

Application Specific Information:
=================================================================
==99820==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000107b28049 bp 0x7ffeecc08710 sp 0x7ffeecc084e0 T0)
==99820==The signal is caused by a READ memory access.
==99820==Hint: address points to the zero page.
    #0 0x107b28048 in QQuickWindowPrivate::renderSceneGraph(QSize const&, QSize const&) (QtQuick:x86_64+0x49e048)
    #1 0x1079911d4 in QSGGuiThreadRenderLoop::renderWindow(QQuickWindow*) (QtQuick:x86_64+0x3071d4)
    #2 0x107b3ee24 in QQuickWindow::event(QEvent*) (QtQuick:x86_64+0x4b4e24)
[...]

PR #588, opened 2020-01-04
Don't clean up scenegraph in RenderModeRequest::apply()

nicolasfella:
cleanupSceneGraph deletes the renderer object in QQuickWindow (qquickwindow.cpp:3729) which leads to a crash in renderSceneGraph() (see #586 )

Removing the cleanup fixes the crash for me, but I don't know about the possible side effects of doing this. At first glance the Quick Inspector seems to work normally


akreuzkamp:
Ok, from all I can tell this is basically a regression in Qt, because at qtquickwindow.cpp:551 they don't check `renderer` for nullness... I intercepted it from the debugger and returned the function whenever renderer is null. This fixes all the issues we have with this.
One way forward would certainly be to just disable custom render modes for Qt 5.14.0 and .1 and fix it in Qt for 5.14.2. I don't see another way currently, but maybe there is?

milianw:
Has anyone fixed this upstream? Is this still needed?

PR #611, opened 2020-09-02
Prefer picking items in fully visible parent-child chains

milianw:
Most notably, don't pick the first item that's visible and has
a non-zero opacity: While the visibility property takes the parent's
visibility into account, that is not the case for the opacity
property. Thus, we need to track that manually, otherwise we
would easily end up picking unrelated items. In the samegame example
e.g. one could easily end up picking an invisible highscore item
in the center of the game canvas - the parent for that item is
visible but has opacity 0. With this patch, we instead pick the game
blocks instead, which is closer to what I'd expect.


milianw:
will merge into 2.11

PR #612, opened 2020-09-21
Make use of all CPUs to parse the processes

bog-dan-ro:
The processList functions it's quite slow when running a lot of processes,
using all the CPUs makes it much faster.

### DIFF
--- a/launcher/ui/processlist.h
+++ b/launcher/ui/processlist.h
@@ -47,8 +47,6 @@ struct ProcData
     QString user;
     GammaRay::ProbeABI abi;
 
-    ProcData() {}
-
     inline bool equals(const ProcData &other) const
     {
         return ppid == other.ppid &&
--- a/launcher/ui/processlist_unix.cpp
+++ b/launcher/ui/processlist_unix.cpp
@@ -35,6 +35,7 @@
 #include <launcher/core/probeabidetector.h>
 
 #include <QProcess>
+#include <QtConcurrent>
 #include <QDir>
 
 #include <algorithm>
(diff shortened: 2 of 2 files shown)


launcher/ui/processlist_unix.cpp
```diff
@@ -127,15 +128,16 @@ ProcDataList processList(const ProcDataList &previous)
     const QStringList procIds = procDir.entryList();
     if (procIds.isEmpty())
         return rc;
-    for (const QString &procId : procIds) {
+    QMutex mutex;
```
milianw, line 131: don't use a mutex, instead actually leverage map:

- let the lambda return ProcData
- return a default-constructed ProcData instead of continue
- filter out all invalid entries once blockingMap is done

milianw (changes requested):

launcher/ui/processlist_unix.cpp
Waqar144: @milianw can you check again?

launcher/ui/processlist_unix.cpp
```diff
@@ -127,18 +128,21 @@ ProcDataList processList(const ProcDataList &previous)
     const QStringList procIds = procDir.entryList();
     if (procIds.isEmpty())
         return rc;
-    for (const QString &procId : procIds) {
+
+    std::function<ProcData(QString)> procIdToProcData = [](const QString &procId) {
```
milianw, line 132: don't force type erasure here, i.e. don't spell out `std::function<...>` - instead use `auto`. I'd personally even go as far as make this a free function in an anon namespace (or `static`) to cleanup the code.

launcher/ui/processlist.h
```diff
@@ -47,7 +47,7 @@ struct ProcData
     QString user;
     GammaRay::ProbeABI abi;
 
-    ProcData() {}
+    ProcData() = default;
```
milianw, line 50: just remove this line altogether (rule of zero)

milianw (changes requested):
yes, that looks much better, a few small nits then this can go in imo

launcher/ui/processlist_unix.cpp
Waqar144: done

launcher/ui/processlist.h
Waqar144: done

PR #622, opened 2021-02-21
MyStrCpy: fixed to work with long typenames

mwu-tow:
I have encountered the crash when inspecting objects with long typenames.
This PR fixes it — `strncpy_s`'s last argument cannot be equal to size, as there would be no space remaining for null character.
(we cannot just copy N characters into N-sized buffer)

Also I see the another line touched, as the file was converted to UTF-8. If you don't want this change, please let me know (or just manually apply the relevant part).


winterz:
we should update 3rdparty/Stackwalker

mwu-tow:
@winterz  Sorry for the delay, I've tried doing as you suggested, the result has been posted as #623. It supersedes this PR.

PR #629, opened 2021-03-22
Build fixes for FreeBSD

unrelentingtech:
This fixes the ABI arch detection and a BFD-related build issue.


winterz:
committed thanks
91917fa608674236150d892cd28297267ae7f2cf

winterz:
FYI: today I updated the 3rdparty/backward-cpp to a more modern version (v1.5) which already had a fix.
hopefully still works for you.

PR #634, opened 2021-05-17
Adding support for Qt for Python applications on Windows

scherzocrk:
This piece of code adds support for PySide or PyQt applications running on Windows. Modifications are described below. Suggestions are more than welcome.

**Attaching:**
- PySide or PyQt creates a .pyd file only on Windows systems named QtCore.pyd which is loaded as a library by python.exe together with Qt5Core.dll. 
- qtCoreForProcess function was finding the .pyd file first and that was a problem. What I did was to add an exclusion based on the import extension
- I tested it and it is working fine

**Launching**
- I had to use a small python snippet to get the correct path of PySide or PyQT files in case virtualenvs are used
- This code finds the correct shared library and gets the right abi information
- I tested and the application is launched but there is a problem because gammaray UI isn't showing (maybe an issue with DLL injecting, not sure). I'd appreciate if you guys could help me out and fix this for a full support (attaching and launching) of Python applications that use Qt

Thanks

### DIFF
--- a/launcher/core/probeabidetector_win.cpp
+++ b/launcher/core/probeabidetector_win.cpp
@@ -40,6 +40,7 @@
 #include <QDir>
 #include <QFile>
 #include <QFileInfo>
+#include <QProcess>
 #include <QStandardPaths>
 #include <QString>
 #include <QStringList>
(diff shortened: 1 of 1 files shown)


MingXingZ:
Hi scherzocrk,  is the problem solved? I can attach to a pyqt application with your code change, but cannot launtch it

PR #639, opened 2021-06-18
Context menu option to copy message stack trace

shaan7:
Useful when you want to share the stack trace for the current message with someone.

### DIFF
--- a/core/stacktracemodel.cpp
+++ b/core/stacktracemodel.cpp
@@ -98,3 +98,17 @@ QVariant StackTraceModel::headerData(int section, Qt::Orientation orientation, i
     }
     return QAbstractTableModel::headerData(section, orientation, role);
 }
+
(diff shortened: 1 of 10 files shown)


core/stacktracemodel.cpp
```diff
@@ -98,3 +98,17 @@ QVariant StackTraceModel::headerData(int section, Qt::Orientation orientation, i
     }
     return QAbstractTableModel::headerData(section, orientation, role);
 }
+
+QStringList StackTraceModel::fullTrace() const
+{
+    QStringList bt;
+    bt.reserve(m_frames.size());
+    for (auto it = m_frames.begin(); it != m_frames.end(); ++it) {
```
dantti, line 106: this could also be a for ranged loop

shaan7: Done :+1:

PR #640, opened 2021-07-06
Launch QtCreator with the -client argument, to reuse the same instance

dfaure-kdab:
Launching a brand new QtCreator each time the user does "Show Source"
is really not convenient. Reusing the same QtCreator even allows to load
the project once and navigate in a proper project after that.


vkrause:
+2

PR #641, opened 2021-07-06
Create QPainter on stack rather than with new.

dfaure-kdab:
This fixes the runtime warning
QBackingStore::endPaint() called with active painter; did you forget to destroy it or call QPainter::end() on it?

@chsterz


vkrause:
+2

PR #667, opened 2021-12-31
Fix reparenting when new parent isn't tracked

Waqar144:
Sometimes on reparent, gammaray doesn't know about the new parent yet.
Ensure to add the parent before reparenting

### DIFF
--- a/core/probe.cpp
+++ b/core/probe.cpp
@@ -846,13 +846,17 @@ bool Probe::eventFilter(QObject *receiver, QEvent *event)
                 // child added events are sent before qt_addObject is called,
                 // so we assumes this comes from the ctor
                 objectAdded(obj, true);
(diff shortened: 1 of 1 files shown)


vkrause (commented):
This doesn't look wrong as such, but doesn't Probe already contain safeguards for exactly that scenario, ie. ensuring parent objects are added before their children? If a scenario is missing there, fixing this in Probe would avoid needing similar workarounds in other tree models I guess?

Waqar144:
> This doesn't look wrong as such, but doesn't Probe already contain safeguards for exactly that scenario, ie. ensuring parent objects are added before their children? If a scenario is missing there, fixing this in Probe would avoid needing similar workarounds in other tree models I guess?

Done.

vkrause:
+2

PR #668, opened 2021-12-31
Don't keep collecting abi info all the time

Waqar144:
The process list hardly changes but we keep doing the same work over and over again. Instead reuse results like we in `unixProcessListPS`

(use `--ignore-space-change` to see the actual diff)

### DIFF
--- a/launcher/ui/processlist_unix.cpp
+++ b/launcher/ui/processlist_unix.cpp
@@ -117,46 +117,63 @@ static ProcDataList unixProcessListPS(const ProcDataList &previous)
     return rc;
 }
 
-static ProcData procIdToProcData(const QString &procId)
+struct ProcIdToProcData
 {
-    ProcData proc;
-    if (!isUnixProcessId(procId))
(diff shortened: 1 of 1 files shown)


launcher/ui/processlist_unix.cpp
```diff
@@ -117,46 +117,65 @@ static ProcDataList unixProcessListPS(const ProcDataList &previous)
...
+    typedef ProcData result_type;
 
-    const QStringList data = QString::fromLocal8Bit(file.readAll()).split(' ');
+    ProcData operator()(const QString &procId) const
+    {
+        ProcData proc;
+        if (!isUnixProcessId(procId))
+            return proc;
+
+        QString filename = QStringLiteral("/proc/");
```
dantti, line 134: while you are at it, also change filename = QLatin1String("/proc/") + procId + QLatin1String("/stat") like you do at 155

Waqar144: done.

dantti (approved):
Works here but `--ignore-space-change` didn't show any diff

PR #670, opened 2022-01-09
Install doc: A couple of updates and a deletion

ferdnyc:
This PR changes the `Install.txt` file (in separate commits) by:

1. Updating two dependencies:
        * The minimum required CMake version is explicitly listed as 3.16.0, to match the `cmake_minimum_version()` call at the top of the `CMakeLists.txt`.
        * Qt 6 support (while arguably implicit in "Qt 5.5 or higher") is made explicit
1. Removing the section at the end of the file that applied to "Qt <= 5.4 only", because 5.5 is the minimum version.


winterz:
thanks

PR #671, opened 2022-01-19
Install doc fixes

krf:
Just noticed that we do not even talk about building GammaRay inside the README.md. This PR fixes it.


winterz (approved):
great, thanks

PR #676, opened 2022-02-03
Upgrade QML Connections {} syntax to functions

ratijas:
Old signal handler declaration style is deprecated and spits warnings at runtime.

### DIFF
--- a/plugins/positioning/mapview.qml
+++ b/plugins/positioning/mapview.qml
@@ -13,7 +13,7 @@
 
 import QtPositioning 5.3
 import QtLocation 5.3
-import QtQuick 2.0
+import QtQuick 2.15
 
 Item {
     Plugin {
@@ -34,7 +34,9 @@ Item {
             opacity: 0.5
             Connections {
                 target: _controller
-                onOverrideCoordinateChanged: sourceTrace.addCoordinate(_controller.sourceCoordinate)
+                function onOverrideCoordinateChanged() {
+                    sourceTrace.addCoordinate(_controller.sourceCoordinate);
+                }
             }
         }
 
@@ -45,7 +47,9 @@ Item {
             opacity: 0.5
             Connections {
                 target: _controller
-                onOverrideCoordinateChanged: overrideTrace.addCoordinate(_controller.overrideCoordinate)
+                function onOverrideCoordinateChanged() {
+                    overrideTrace.addCoordinate(_controller.overrideCoordinate);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


vkrause (commented):
This will likely break backward compatibility with older Qt5 versions, wont it? (GammaRay supports Qt5 back to 5.5).

ratijas:
> This will likely break backward compatibility with older Qt5 versions, wont it? (GammaRay supports Qt5 back to 5.5).

5.5?  oh.  Is that something that still exists?

Likely will break than. But the current approach is surely broken for Qt 6, isn't it?

vkrause:
QtLocation is gone in Qt6, so this entire thing doesn't work there anyway.

ratijas:
At least Widget Inspector is still legit, I guess

ratijas:
> Personally I think it would be fine to upgrade to a newer Qt version nowadays, no?

Too early. The KDE  Frameworks are not yet fully ported to Qt6, so application styling UX would be rather suboptimal on Linux platform.

> If people want a gammaray that works against Qt 5.5, they can use an older release.

So a GammaRay build against Qt6 could still inspect apps running on Qt5? That would be great. Never tried (I think?)

ratijas:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/KDAB/GammaRay?pullRequest=676) before we can accept your contribution.

I did, some time ago. And my contributions were already merged in GammaRay in past.

dantti:
> So a GammaRay build against Qt6 could still inspect apps running on Qt5? That would be great. Never tried (I think?)

No, that's not possible because GammaRay injects itself into the program, this is even a problem for same major version but different minor ie Qt6.2 GammaRay vs Qt6.4 App, for best compat you should use that exact same version, so I agree with @milianw that we could drop 5.5, but iirc this warning came around 5.14 or so...
5.14 indeed according to https://doc.qt.io/archives/qt-5.14/qml-qtqml-connections.html

redstrate:
@ratijas Will need a manual rebase now, looks like some stuff was shuffled.

dantti:
@ratijas now that min required version is 5.15 could you please rebase this?

ratijas:
no problem!

ratijas:
image

AAAAAaaaa how is this a valid syntax

ratijas:
Widget inspector was removed, so it's not relevant anymore.

PR #682, opened 2022-02-17
Support Qt6

Waqar144:


vkrause:
This seems to contain a few commits from me that were never intended for integration, but as temporary placeholders, ie. changes that should be replaced by proper solutions, not built on top of. Is this something we can still clean up a bit?

Waqar144:
> This seems to contain a few commits from me that were never intended for integration, but as temporary placeholders, ie. changes that should be replaced by proper solutions, not built on top of. Is this something we can still clean up a bit?

Yes, i have created the PR to start preparing it for master.

CMakeLists.txt
```diff
@@ -397,6 +406,17 @@ if(GAMMARAY_BUILD_DOCS)
   find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS Help LinguistTools)
 endif()
 
+if(${Qt_VERSION_MAJOR} GREATER_EQUAL 6)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  add_definitions(-DGAMMARAY_QT6_TODO)
```
vkrause, line 412: Do we still need this? That was supposed to be just for temporary scaffolding during porting.

cmake/ECMQueryQmake.cmake
```diff
@@ -0,0 +1,58 @@
+#
```
vkrause, line 1: This file seems unused? There's also one with the same name in cmake/ECM/modules (with a removed copyright header and using a non-existent include?). Maybe some kind of merge conflict?

core/tools/messagehandler/messagehandler.cpp
```diff
@@ -149,10 +149,10 @@ MessageHandler::MessageHandler(Probe *probe, QObject *parent)
     // install handler directly, catches most cases,
     // i.e. user has no special handler or the handler
     // is created before the QApplication
-    ensureHandlerInstalled();
+//     ensureHandlerInstalled();
     // recheck when eventloop is entered, if the user
     // installs a handler after QApp but before .exec()
-    QMetaObject::invokeMethod(this, "ensureHandlerInstalled", Qt::QueuedConnection);
+//     QMetaObject::invokeMethod(this, "ensureHandlerInstalled", Qt::QueuedConnection);
```
vkrause, line 155: Are these changes intentional or should this be conditional on the Qt version?

examples/quick-batching/CMakeLists.txt
```diff
@@ -1 +1 @@
-qml_lint(quick-batching.qml Slider.qml)
+#qml_lint(quick-batching.qml Slider.qml)
```
vkrause, line 1: Temporary porting leftover? If that doesn't work with 6 we probably should at least make it conditional on the Qt version. Repeats below.

plugins/quickinspector/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ if(TARGET Qt::Quick)
       SOURCES ${gammaray_quickinspector_srcs}
     )
 
-    if(NOT Qt5Quick_VERSION VERSION_LESS 5.7)
+    if(NOT Qt5Quick_VERSION VERSION_LESS 5.7 OR TARGET Qt6::Quick)
       target_sources(gammaray_quickinspector PUBLIC ${CMAKE_CURRENT_LIST_DIR}/quickimplicitbindingdependencyprovider.cpp)
```
vkrause, line 44: There's build fixes for this file below, but it's taken out for Qt6 here it seems?

examples/quick-batching/CMakeLists.txt
Waqar144: done. examples seem to work, so renabled.

cmake/ECMQueryQmake.cmake
Waqar144: yes, removed this one as it was recently moved into `cmake/ECM` directory.

plugins/quickinspector/CMakeLists.txt
Waqar144: rather its being added explicitly for Qt6

core/tools/messagehandler/messagehandler.cpp
Waqar144: done.

vkrause:
Let's get this in then.

PR #683, opened 2022-02-17
WidgetInspector: Auto expand the tree on filtering

Waqar144:
When you are filtering for a particular widget, Gammaray doesn't expand
the tree for you. This can be really annoying when you have long object
hierarchies as you need to manually expand each node to get to the item
you want. To fix this, auto expand the tree after filtering.

### DIFF
--- a/plugins/widgetinspector/widgetinspectorwidget.cpp
+++ b/plugins/widgetinspector/widgetinspectorwidget.cpp
@@ -44,6 +44,7 @@
 
 #include "common/objectbroker.h"
 #include "common/objectmodel.h"
+#include "common/remotemodelroles.h"
 
 #include <ui/contextmenuextension.h>
 #include <ui/paintbufferviewer.h>
@@ -61,6 +62,7 @@
 #include <QSettings>
 #include <QLayout>
 #include <QTabBar>
+#include <QTimer>
 
 using namespace GammaRay;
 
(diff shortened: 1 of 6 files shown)


plugins/widgetinspector/widgetinspectorwidget.cpp
```diff
@@ -198,6 +200,90 @@ void WidgetInspectorWidget::onTabChanged(int index)
...
+
+    const int rowCount = model->rowCount(idx);
+    for (int i = 0; i < rowCount; ++i) {
+        auto childIdx = model->index(i, 0, idx);
+        if (!childIdx.isValid()) {
+            continue;
+        }
+
+        // The value might not be there, store it for delayed expansion
+        if (childIdx.data().toString() == s_emptyDisplayValue) {
```
vkrause, line 279: There is RemoteModelRole::LoadingState, which should allow for a more robust and efficient check for this I think.

vkrause (commented):
Good idea! Feels like something that could be generic and reused in a number of other views as well? That can be a subsequent improvement of course.

Waqar144:
> Good idea! Feels like something that could be generic and reused in a number of other views as well? That can be a subsequent improvement of course.

Done. I have moved the expansion logic to SearchLineController. Once the search finishes (user stops typing), it tries to expand the nodes recursively. For now, I have enabled this for only 3 treeviews `(objectInspector, widgetInspector, quickInspector)`

@vkrause ^

vkrause (approved):

PR #684, opened 2022-02-18
Fix a lot of clang-tidy readibility warnings

Waqar144:
When writing code especially using clangd, there are a lot of warnings which make it hard to focus. This change addresses most of these.

### DIFF
--- a/client/clientconnectionmanager.h
+++ b/client/clientconnectionmanager.h
@@ -69,7 +69,7 @@ public:
     void connectToHost(const QUrl &url, int tryAgain = 0);
 
     /*! Manually show the splash screen. */
-    void showSplashScreen();
+    static void showSplashScreen();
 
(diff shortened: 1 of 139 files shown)


vkrause:
None of this seems to change functionality, so no objection from my side. 

If this is only about getting rid of the warnings rather than actually wanting this style change (particularly looking at the qualified-auto change), I'm sure there's ways to disable those by other means? I don't have a preference either way, so pick whatever you prefer.

Waqar144:
> None of this seems to change functionality, so no objection from my side.
> 
> If this is only about getting rid of the warnings rather than actually wanting this style change (particularly looking at the qualified-auto change), I'm sure there's ways to disable those by other means? I don't have a preference either way, so pick whatever you prefer.

Dropped the qualified-auto commit as it is now being ignored via the `.clangd` file.

PR #689, opened 2022-04-14
In Positioning plugin, change Qt version check #ifdefs around changed…

reedhedges:
In Positioning plugin (plugins/positioning/geopositioninfosourcefactory.cpp), change Qt version check #ifdefs around use of QFactoryLoader metadata object from >= 6.4.0, to >= 6.3.0 instead.

Fixes this compile error when using Qt 6.3.0:
````
plugins/positioning/geopositioninfosourcefactory.cpp:99:50: error: cannot convert ‘QString’ to ‘QtPluginMetaDataKeys’
````

I couldn't find any documentation of exactly what Qt release this change may have first appeared in, I just tested with 6.2.3 and 6.3.0.

### DIFF
--- a/plugins/positioning/geopositioninfosourcefactory.cpp
+++ b/plugins/positioning/geopositioninfosourcefactory.cpp
@@ -63,7 +63,7 @@ QGeoPositionInfoSource* GeoPositionInfoSourceFactory::positionInfoSource(QObject
 
     // filter anything not applicable
     for (auto it = indexes.begin(); it != indexes.end();) {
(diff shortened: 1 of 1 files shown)


winterz:
thank you

reedhedges:
Is this generally the right branch to keep using if using Qt 6; is the idea
that it will merge into master at some point, or are you keeping qt5/qt6
separate for a while? (maybe as long as qt5 is supported?)

winterz:
no, use master from now on.
I'm going to delete the work/qt6-support branch

PR #697, opened 2022-05-15
CMakeLists.txt: Move Qt version check downwards

### LINKED ISSUE #696: 04afc773e3c9779f85607e0a1ffa0ac7c94caae6 breaks CMake configure
```
CMake Error at CMakeLists.txt:291 (if):
  if given arguments:

    "GREATER_EQUAL" "6"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
See also "CMakeFiles/CMakeOutput.log".
See also "CMakeFiles/CMakeError.log".
```

This is because at the time the check is performed, the Qt libraries have not been found yet with `find_package`. While it is true that `-Wa-mbig-obj` is only needed with Qt6, the check in its current form breaks configuring for both Qt5 and Qt6. So the check needs to be moved further down.

JohannesKauffmann:
Previously, the Qt version was checked before find_package, resulting
in a broken CMake configure on MinGW regardless of Qt version.

Fixes #696.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,10 +288,6 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   if(MINGW)
     # mingw will error out on the crazy casts in probe.cpp without this
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fpermissive")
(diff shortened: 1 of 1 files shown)


JohannesKauffmann:
Ah, well, in hindsight, your fix is better anyway, since now you make use of a Qt version check that was already there. Thanks for merging quickly!

Issue #655, opened 2021-10-23
Unable to attach to a running application


thawfeekyahya:
Hi ,

I am trying to attach gammray to running qt application, but gammaray fails to attach and the application crashes.

Gammaray and the application both are compiled with same Qt version.

This is the error from Gammaray 
`cannot call method fatalMessageReceived on unknown object of name com.kdab.GammaRay.MessageHandler with address 71 - did you forget to register it?
Cannot dispatch message 31 - no handler registered. Receiver: com.kdab.GammaRay.MessageHandler, address 71
`

and this back trace log from application after the crash,

`QFatal in Qt Application Manager (/home/osboxes/qt_path/bin/appman)
START BACKTRACE:
1	QMessageLogger::fatal(char const*, ...) const ()
2	/home/osboxes/qt_path/lib/libQt5Core.so.5 ()
3	QMetaType::registerNormalizedType(QByteArray const&, void (*)(void*), void* (*)(void*, void const*), int, QFlags<QMetaType::TypeFlag>, QMetaObject const*) ()
4	/home/osboxes/build_Gammaray/lib/libgammaray_core-qt5_15-x86_64.so.2.11.1 ()
5	GammaRay::AggregatedPropertyModel::rowCount(QModelIndex const&) const ()
6	/home/osboxes/build_Gammaray/lib/libgammaray_core-qt5_15-x86_64.so.2.11.1 ()
7	QMetaMethod::invoke(QObject*, Qt::ConnectionType, QGenericReturnArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument) const ()
8	GammaRay::Endpoint::dispatchMessage(GammaRay::Message const&) ()
9	GammaRay::Endpoint::readyRead() ()
10	/home/osboxes/qt_path/lib/libQt5Core.so.5 ()
11	QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
12	QIODevice::readyRead() ()
13	/home/osboxes/qt_path/lib/libQt5Network.so.5 ()
14	/home/osboxes/qt_path/lib/libQt5Network.so.5 ()
15	/home/osboxes/qt_path/lib/libQt5Network.so.5 ()
16	/home/osboxes/qt_path/lib/libQt5Network.so.5 ()
17	/home/osboxes/qt_path/lib/libQt5Network.so.5 ()
[...]

PR #702, opened 2022-06-09
Fix bug wit unable to attach to a running application. (#655)

bohdansyrovetnyk:
It's a problem related to `QHash` since it sometimes makes rehash when we insert a new value. Since `addPropertyAdaptor` updates `m_parentChildrenMap` we shouldn't save the reference before inserting new data because after rehash the reference becomes invalid.

### DIFF
--- a/core/aggregatedpropertymodel.cpp
+++ b/core/aggregatedpropertymodel.cpp
@@ -101,7 +101,7 @@ void AggregatedPropertyModel::clear()
     if (!m_rootAdaptor)
         return;
 
-    const auto count = m_parentChildrenMap.value(m_rootAdaptor).size();
+    const auto count = m_parentChildrenMap.at(m_rootAdaptor).size();
     if (count)
         beginRemoveRows(QModelIndex(), 0, count - 1);
 
@@ -304,7 +304,7 @@ int AggregatedPropertyModel::rowCount(const QModelIndex &parent) const
     if (!m_rootAdaptor || parent.column() > 0)
         return 0;
     if (!parent.isValid())
-        return m_parentChildrenMap.value(m_rootAdaptor).size();
+        return m_parentChildrenMap.at(m_rootAdaptor).size();
 
     auto adaptor = adaptorForIndex(parent);
     auto &siblings = m_parentChildrenMap[adaptor];
@@ -324,7 +324,7 @@ int AggregatedPropertyModel::rowCount(const QModelIndex &parent) const
     auto childAdaptor = siblings.at(parent.row());
     if (!childAdaptor)
         return 0;
-    return m_parentChildrenMap.value(childAdaptor).size(); // childAdaptor->count() might already be updated in insert/removeRows
+    return m_parentChildrenMap.at(childAdaptor).size(); // childAdaptor->count() might already be updated in insert/removeRows
 }
 
 Qt::ItemFlags AggregatedPropertyModel::flags(const QModelIndex &index) const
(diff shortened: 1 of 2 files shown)


bohdansyrovetnyk:
@Waqar144, please review

Waqar144:
The changes seems okay however I think it would be better if we used `std::unordered_map` here instead. Can you do that?

bohdansyrovetnyk:
@Waqar144l, I will do it but the current fix is also needed in the case of `std::unordered_map` since the documentation said `"If an insertion occurs and results in a rehashing of the container, all iterators are invalidated. Otherwise, iterators are not affected. References are not invalidated. Rehashing occurs only if the new number of elements is greater than max_load_factor()*bucket_count()."`

Waqar144:
> @Waqar144l, I will do it but the current fix is also needed in the case of `std::unordered_map` since the documentation said `"If an insertion occurs and results in a rehashing of the container, all iterators are invalidated. Otherwise, iterators are not affected. References are not invalidated. Rehashing occurs only if the new number of elements is greater than max_load_factor()*bucket_count()."`

Hmm, right. Then let's have this for now and we can think about using std::unordered_map later.

Waqar144:
> I will do it but the current fix is also needed in the case of std::unordered_map

Actually looking more closely at the doc, it says:

> References are not invalidated.

So IIUC the current fix won't be needed.

bohdansyrovetnyk:
@Waqar144, please review one more time

core/aggregatedpropertymodel.cpp
```diff
@@ -475,22 +477,23 @@ bool AggregatedPropertyModel::hasLoop(PropertyAdaptor *adaptor, const QVariant &
...
+    auto oldAdaptor = m_parentChildrenMap.at(parentAdaptor).at(index);
     if (oldAdaptor) {
-        auto oldRowCount = m_parentChildrenMap.value(oldAdaptor).size();
+        auto oldRowCount = m_parentChildrenMap.at(oldAdaptor).size();
         if (oldRowCount > 0)
             beginRemoveRows(createIndex(index, 0, parentAdaptor), 0, oldRowCount - 1);
         m_parentChildrenMap[parentAdaptor][index] = nullptr;
-        m_parentChildrenMap.remove(oldAdaptor);
+
+        m_parentChildrenMap.erase(m_parentChildrenMap.find(oldAdaptor));
```
Waqar144, line 496: double lookup not needed I think, iterator from the line `auto oldRowCount = m_parentChildrenMap.at(oldAdaptor).size();` can be used here.

core/aggregatedpropertymodel.cpp
```diff
@@ -304,10 +304,10 @@ int AggregatedPropertyModel::rowCount(const QModelIndex &parent) const
     if (!m_rootAdaptor || parent.column() > 0)
         return 0;
     if (!parent.isValid())
-        return m_parentChildrenMap.value(m_rootAdaptor).size();
+        return m_parentChildrenMap.at(m_rootAdaptor).size();
 
     auto adaptor = adaptorForIndex(parent);
-    auto &siblings = m_parentChildrenMap[adaptor];
+    auto &siblings =  m_parentChildrenMap[adaptor];
```
Waqar144, line 310: uneeded space

Waqar144 (changes requested):

core/aggregatedpropertymodel.cpp
bohdansyrovetnyk: Fixed

core/aggregatedpropertymodel.cpp
bohdansyrovetnyk: Fixed

PR #717, opened 2022-07-24
Ignore event dispatcher signals

Waqar144:
There can be thousands of signals coming in from EventDispatchers and they are not useful for a normal app developer.


winterz:
lgtm

PR #719, opened 2022-07-24
Avoid fetching Declaration/Creation location for every index

Waqar144:
Fetching them can be really expensive


vkrause:
IIRC this role is accessed synchronously from the context menu, if that still behaves correctly with the delayed availability this looks ok to me.

PR #720, opened 2022-07-25
Flatten ObjectTreeModel::indexForObject

Waqar144:
Less lookups and avoid the cost of setting up stack for every recursive
call.

### DIFF
--- a/core/objecttreemodel.cpp
+++ b/core/objecttreemodel.cpp
@@ -260,15 +260,14 @@ QModelIndex ObjectTreeModel::indexForObject(QObject *object) const
 {
     if (!object)
         return {};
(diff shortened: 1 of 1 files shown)


vkrause:
Do we even need to walk through the entire chain here? It looks like all we need is the list of siblings, which we can get from the internal maps, no? (The reason/mistake in the old code seems to be using index() rather than createIndex(), which made that necessary).

Waqar144:
> Do we even need to walk through the entire chain here? It looks like all we need is the list of siblings, which we can get from the internal maps, no? (The reason/mistake in the old code seems to be using index() rather than createIndex(), which made that necessary).

Done

PR #721, opened 2022-07-25
Rate limit data changed signals

Waqar144:
Hand picked a couple of commits from PR #377 which can go in imo.

### DIFF
--- a/core/remote/remotemodelserver.cpp
+++ b/core/remote/remotemodelserver.cpp
@@ -44,6 +44,7 @@
 #include <QIcon>
 #include <QSequentialIterable>
 #include <QSortFilterProxyModel>
+#include <QTimer>
 
 #include <iostream>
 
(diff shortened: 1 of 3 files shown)


vkrause:
I think moving this to the RemoteModel makes sense, we have a few models doing this on their own IIRC (something we can maybe remove following this). For extra safety we might want to flush the pending data change signals before sending any other model event to the client (inserts, removals, etc), the whole remote model thing is based on the invariant that things arrive in order. I could imagine client-side filtering/sorting to get confused otherwise.

core/remote/remotemodelserver.cpp
```diff
@@ -339,9 +349,33 @@ void RemoteModelServer::dataChanged(const QModelIndex &begin, const QModelIndex
...
+            } else if (!change.roles.isEmpty()) {
+                foreach (const int role, roles) {
+                    if (!change.roles.contains(role)) {
+                        change.roles << role;
+                    }
+                }
+            }
+        } else {
+            m_pendingDataChanged.insert(index, DataChanged(begin.column(), end.column(), roles));
+        }
```
ferdnyc, line 374:
```suggestion
        if (it == m_pendingDataChanged.end()) {
            m_pendingDataChanged.insert(
                index, DataChanged(begin.column(), end.column(), roles));
            continue;
        }

        DataChanged &change = it.value();

        change.left = qMin(change.left, begin.column());
        change.right = qMax(change.right, end.column());

        if (roles.isEmpty()) {
            change.roles.clear();
        } else if (!change.roles.isEmpty()) {
            foreach (const int role, roles) {
                if (!change.roles.contains(role)) {
                    change.roles << role;
                }
            }
        }
```

PR #730, opened 2022-09-17
Fix warnings

JohannesKauffmann:
- Fixed warning about empty string format specifier
- Use nullptr instead of 0


Waqar144 (approved):

winterz:
i'm not so sure about using nullptr in those windows system calls
my memory on this is hazy, but I'm pretty sure you don't want to those changes

JohannesKauffmann:
After all the Windows ifdefery and macro replacement is said and done by the preprocessor, the parameters types of these Windows system calls are either a pointer or not. E.g.: `LPCSTR` is really a `const char*`. So using `nullptr` should be fine.

If `nullptr` is not wanted in this case, at least those cases where the type is a pointer should be changed to `NULL` instead of `0`, since currently, `0` is used both for pointer types (like `LPCSTR`) and value types (like `DWORD`).

Waqar144:
As long as the type is a pointer, using `nullptr` should be okay, iirc nullptr can implicitly convert to NULL anyways.

JohannesKauffmann:
Resolved the conflicts caused by recent changes.

PR #741, opened 2022-11-26
Pan with Middle mouse button in remote view

MiKom:
Allow panning with middle mouse button regardless of the current interaction mode.

This reduces the need to switch the interaction mode in the remote view when the area where the user wants to look at is out of the view.


MiKom:
Converting to draft to include the input after fruitful discussion with @Waqar144 :
 * Try some different key combination for panning to make it usable on touchpads that don't have middle mouse button
 * add some information about it to pan tool tooltip

Issue #742, opened 2022-11-27
Updated ECM dependencies require use of BUILD_WITH_QT6 option (instead of / in addition to GAMMARAY_QT6_BUILD)


ferdnyc:
With #740 fixed and the `QtVersionOption.cmake` file now present in the repo, it appears that the option it provides, `BUILD_WITH_QT6`, **must** be enabled when configuring to build the Qt6 version of GammaRay, if Qt5 is also present on the build system. (Or, the `QT_MAJOR_VERSION` variable can be manually set to `6`, which is effectively what happens when the `BUILD_WITH_QT6` option is enabled.)

Regardless of the setting of other options (like `GAMMARAY_QT6_BUILD`), or the placement of Qt6 paths in the usual locations (`CMAKE_PREFIX_PATH` / the `PATH` environment variable, etc.), if `QT_MAJOR_VERSION` is not defined and `BUILD_WITH_QT6` is not set, `QtVersionOption.cmake` will default to configuring for Qt5. This can lead to difficult-to-diagnose errors when running CMake.

(That's an understatement. The errors are, quite frankly, bizarre. For example, the **Qt6 CMake configuration** might error out with a complaint that `qt_generate_foreign_qml_types() is only available in Qt 6`:

```console
$ cmake -B x86_64-redhat-linux-gnu_qt6 -S . -DGAMMARAY_QT6_BUILD:BOOL=TRUE                               
-- The CXX compiler identification is GNU 12.2.1
-- The C compiler identification is GNU 12.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Setting build type to Release as none was specified.
-- Found libdw: /usr/lib64/libdw.so  
-- Found libbfd: /usr/lib64/libbfd.so  
-- Could NOT find libdwarf (missing: LIBDWARF_LIBRARY LIBDWARF_INCLUDE_DIR) 
-- Found Backward: ...GammaRay/3rdparty/backward-cpp  
-- Building GammaRay 2.99.50 in Release mode
-- Performing Test C_SUPPORTS_UNUSED_BUT_SET
[...]

PR #743, opened 2022-11-27
Fix configuration of Qt6 builds

ferdnyc:
- Delay include of `ECMGeneratePriFile.cmake` until after Qt configuration is completed
- Set the `QT_MAJOR_VERSION` variable used by ECMQueryQt.cmake

Fixes: #742


ferdnyc:
Ugh. Now I'm not so sure about this change. Not that I think it's not correct, but I think it may not be enough. And the previous brokenness may have been having some beneficial effects.

On Fedora 37, it does the right thing and causes the build to successfully configure when `GAMMARAY_QT6_BUILD` is set true, whereas without this change setting that option results in the error I reported in #742.

But on Fedora 36, with this change (**or without it**, as it turns out), for reasons I don't totally understand the build is defaulting to Qt6 unless `-DQt_DIR=/usr/lib64/cmake/Qt5` is set on the `cmake` command line.

And even on Fedora 37, when a Qt6 build is performed the `Qt_DIR` cache variable will be set to `/usr/lib64/cmake/Qt5` due to the initial versionless Qt probe, which seems like a recipe for problems. (`Qt6_DIR` _is_ set to `/usr/lib64/cmake/Qt6`, at least.)

...The way the `CMakeLists.txt` is overriding the value of `Qt_VERSION_MAJOR` _after_ the initial Qt package probe just doesn't feel quite right to me.

Issue #716, opened 2022-07-22
probeabidetector_mac needs support for FAT libraries


tsdgeos:
Currently only supports MH_MAGIC and MH_MAGIC_64 the Qt6 libraries have a FAT_CIGAM marker.

See https://opensource.apple.com/source/xnu/xnu-792/EXTERNAL_HEADERS/mach-o/fat.h.auto.html problem is that given that one library can have multiple abis now, so i guess that abiForExecutable should return a list of ProbeABI instead of just one?

PR #746, opened 2022-11-30
Make it possible to start Qt 6 Mac applications

tsdgeos:
Qt 6 libs are now fat libraries, i.e. they contain both the amd64 and the arm64 libraries inside, so we need a way to say that a given Qt supports more than one architecture.


Waqar144:
Thanks!

Does this close https://github.com/KDAB/GammaRay/issues/716?

tsdgeos:
Yes, it does

PR #752, opened 2023-01-03
Add LSP/clangd stuff to gitignore

ratijas:
Useful for kdesrc-build setup as well as for other forms of linking compile_commands.json into source directory.


Waqar144:
Thank you

PR #753, opened 2023-01-04
Add zsh completions (compdef) for gammaray

ratijas:
This patch brings zsh completions for every option documented in `--help` output of GammaRay.

image

image

Features:
- [x] completions for probes with displayed human-readable descriptions while completing actual names;
- [x] basic completions for probes (there's no CLI to fetch list of valid probes for current platform though)
- [x] listen & connect addresses are handled separately, for better or worse;
- [x] `--pid`, `--connect` & `<application> <args>` are actually mutually exclusive. Typing one onto the command line prevents others from being considered for completions.
- [x] `<args>` are completed as normal arguments for any "completable"  `<application>`.

I might need help with CMake stuff for cross-platform. Specifically, I have no idea where to install this compdef on OS X.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -732,6 +732,8 @@ else()
         CACHE PATH "Install location of Qt Assistant help files."
     )
     gammaray_convert_to_relative_path(QCH_INSTALL_DIR)
+    set(ZSHAUTOCOMPLETE_INSTALL_DIR "${DATAROOTDIR}/zsh/site-functions")
+
     if(WIN32)
         set(PLUGIN_INSTALL_DIR "plugins/gammaray")
         set(LIBEXEC_INSTALL_DIR "${BIN_INSTALL_DIR}")
--- a/launcher/cli/CMakeLists.txt
+++ b/launcher/cli/CMakeLists.txt
@@ -32,6 +32,13 @@ gammaray_set_rpath(gammaray ${BIN_INSTALL_DIR})
 install(
     TARGETS gammaray ${INSTALL_TARGETS_DEFAULT_ARGS}
 )
+if(ZSHAUTOCOMPLETE_INSTALL_DIR)
+    install(
+        FILES completions/gammaray.zsh
+        RENAME _gammaray
+        DESTINATION ${ZSHAUTOCOMPLETE_INSTALL_DIR}
+    )
+endif()
 if(MSVC)
     install(
         FILES "$<TARGET_PDB_FILE_DIR:gammaray>/$<TARGET_PDB_FILE_NAME:gammaray>"
--- a/launcher/cli/completions/gammaray.zsh
+++ b/launcher/cli/completions/gammaray.zsh
@@ -1,0 +1,90 @@
+#compdef gammaray
+
+# This file is part of GammaRay, the Qt application inspection and manipulation tool.
+#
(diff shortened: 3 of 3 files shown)


ratijas:
where should i put the copyright text?  in the completions file itself or in reuse file?  and should i put my name there, or just copy the KDAB text from other files?

Waqar144:
The copyright text will be at the top of the file. You can add yourself as the author. See any other cpp/h file in the repo. Also please check out https://github.com/KDAB/GammaRay/wiki/Get-Involved, we will need a CAA before proceeding.

ratijas:
> The copyright text will be at the top of the file. You can add yourself as the author. See any other cpp/h file in the repo. Also please check out [KDAB/GammaRay/wiki/Get-Involved](https://github.com/KDAB/GammaRay/wiki/Get-Involved), we will need a CAA before proceeding.

That Copyright Assignment Agreement link needs to be updated. I guess [this is it](https://github.com/KDAB/GammaRay/blob/master/docs/GammaRay-CopyrightAssignmentForm.pdf), but idk how to send an MR to Wiki though

ratijas:
The form pdf is a bit buggy, by the way. It forces some fields to be linked and use the same text, well, at least in Firefox. I can't type in the list of files — it just copies one line seven times like this:

image
image

ratijas:
Oh, turns out, there's also an undocumented `--no-inprocess` argument. Might be worth including as well.

winterz:
> > The copyright text will be at the top of the file. You can add yourself as the author. See any other cpp/h file in the repo. Also please check out [KDAB/GammaRay/wiki/Get-Involved](https://github.com/KDAB/GammaRay/wiki/Get-Involved), we will need a CAA before proceeding.
> 
> That Copyright Assignment Agreement link needs to be updated. I guess [this is it](https://github.com/KDAB/GammaRay/blob/master/docs/GammaRay-CopyrightAssignmentForm.pdf), but idk how to send an MR to Wiki though

I fixed that link

ratijas:
> Oh, turns out, there's also an undocumented `--no-inprocess` argument. Might be worth including as well.

I think it's a leftover from some refactoring in past, but `--no-inprocess` option isn't used by anything. It's only being set in `LaunchOptions::execute`, and never parsed back.

Speaking of CCA. I don't have any kind of digital signature, and I don't think I could get a suitable one. So, printing out those last 3 pages, filling in with pen on paper, scanning back, and sending pictures combined in pdf to info@ email would be OK?

ratijas:
Added copyright/license text; squashed pre-commit.ci formatting changes; sent an email with signed CCA to KDAB.

krf:
Heya, I don't know how to write ZSH completion definitions myself, but was asked to test this since I'm a vivid ZSH user.

The completion seems to work reliably well, but I noticed one thing: After `gammaray --help` the compdef still allows me to  specify an external command; shouldn't it say "no more arguments" there?

Other than that this looks like a super useful addition, thanks!

ratijas:
Hi Kevin, thanks for testing it out!

You are right about `--help` dead-end stuff not stopping external commands from completing anyway. At the time of writing I couldn't make it work, no matter how I tried to combine those options and '(exclusion lists)'. But now that you mentioned it, `cp`, `losetup` and other commands handle it correctly. So I followed their pattern, and now gammaray works better: it won't complete after `--help` &amp; `--version`, but on the other hand it won't suggest them after `--pid` &amp; `--connect` options as they are in a mutually exclusive sets with help options now. I hope this helps!

PR #759, opened 2023-02-16
Show potential launcher errors in a message box

redstrate:
A common thing I run into is trying to launch GammaRay and when injecting, it suddenly closes. This is because injector errors are printed into console instead of the gui, even when launched via it. These are the same potential launcher errors found under Self-Test, but those are easily missable. Now, if built with Qt Widgets - the injector will display a message box showing the errors it has collected:

image

To do this, I made `Launcher::injectorError()` stack every error it has collected, and moved around some of the logging (like printing out potential launcher errors, because calling `injectorError()` will handle that for us). I'm not too happy with copy & pasting around the troubleshooting lines (because `injectorError()` could be called multiple times), but I'm not sure if it's worth adding a new function for.

### DIFF
--- a/launcher/cli/main.cpp
+++ b/launcher/cli/main.cpp
@@ -32,6 +32,7 @@
 
 #ifdef HAVE_QT_WIDGETS
 #include <QApplication>
+#include <QMessageBox>
 #else
 #include <QCoreApplication>
 #endif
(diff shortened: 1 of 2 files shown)


Waqar144:
Hi, thanks for the patch.

Can you please check out https://github.com/KDAB/GammaRay/wiki/Get-Involved, we will need a CAA before proceeding

redstrate:
> Hi, thanks for the patch.
> 
> Can you please check out https://github.com/KDAB/GammaRay/wiki/Get-Involved, we will need a CAA before proceeding

Sent the CAA to info@kdab.com now :)

PR #765, opened 2023-03-09
Fix GAMMARAY_CLIENT_ONLY_BUILD

### LINKED ISSUE #766: GAMMARAY_CLIENT_ONLY_BUILD breaks in multiple places
Attempting a build configured for client-only builds currently results in the following error (if `Qt6::Qml` is available):

```console
$ cmake -B _build -S . -DGAMMARAY_QT6_BUILD=1 -DGAMMARAY_CLIENT_ONLY_BUILD=1 ...
[...]
CMake Error at plugins/qmlsupport/CMakeLists.txt:42 (target_sources):
  Cannot specify sources for target "gammaray_qmlsupport" which is not built
  by this project.
```

This is due to a premature `endif()` in `plugins/qmlsupport/CMakeLists.txt`; the `gammaray_qmlsupport` target is not created in a `CLIENT_ONLY` config, but the code still attempts to set sources on it.

In addition, `tests/CMakeLists.txt` twice calls `gammaray_add_probe_test()` without first checking for `GAMMARAY_CLIENT_ONLY_BUILD`. Seems like a `GAMMARAY_CLIENT_ONLY_BUILD` run needs to be part of the CI jobs, to catch these before they're committed.

Fixed by #765

ferdnyc:
The `target_sources(gammaray_qmlsupport ...)` call needs to be inside the `if(NOT GAMMARAY_CLIENT_ONLY_BUILD)`, because that target won't be defined during a client-only build.

Currently, attempting a `GAMMARAY_CLIENT_ONLY_BUILD` on Qt6 with `Qt6::Qml` available results in:

```text
CMake Error at plugins/qmlsupport/CMakeLists.txt:42 (target_sources):
  Cannot specify sources for target "gammaray_qmlsupport" which is not built
  by this project.
```

Fixes #766

### DIFF
--- a/plugins/qmlsupport/CMakeLists.txt
+++ b/plugins/qmlsupport/CMakeLists.txt
@@ -33,10 +33,11 @@ if(NOT GAMMARAY_CLIENT_ONLY_BUILD)
         Qt::Qml
         Qt::QmlPrivate
     )
-endif()
 
-if(NOT QtQml_VERSION VERSION_LESS 5.10 OR TARGET Qt6::Qml)
-    target_sources(gammaray_qmlsupport PUBLIC qmlbindingprovider.cpp)
+    if(NOT QtQml_VERSION VERSION_LESS 5.10 OR TARGET Qt6::Qml)
+        target_sources(gammaray_qmlsupport PUBLIC qmlbindingprovider.cpp)
+    endif()
+
 endif()
 
 # ui plugin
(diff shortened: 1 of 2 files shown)


ferdnyc:
I also found two instances of `tests/CMakeLists.txt` calling `gammaray_add_probe_test()` without first checking for `GAMMARAY_CLIENT_ONLY_BUILD`); both are corrected by the second commit here.

PR #768, opened 2023-03-11
Add Github Actions CI builds

ferdnyc:
After discovering that `GAMMARAY_CLIENT_ONLY_BUILD`s are broken in the repo (#766, fix proposed in #765), I realized that there were no CI **builds** being done (which would've caught that).

(There are buildbot CI jobs being run against the repo, but they're not public. So for anyone outside of KDAB, there's no way to know _what_ they test. Presumably, not client-only builds! And also not Qt6 builds, based on the job names.)

So, this PR adds Github Actions configuration to run a set of GammaRay build and install tests on every push to, and pull request against, the `master` branch.

- Both `ubuntu-latest` and `macos-latest` platforms are tested
- Two versions of Qt are tested: `5.15.*` and `6.4.*`
- For each os+Qt, two builds are done:
  1. Using default options
  2. Configured with `-DGAMMARAY_CLIENT_ONLY_BUILD=1`

These CI builds are **EXPECTED TO FAIL**, currently. This PR depends on my PR #765, which will fix client-only builds in the `master` branch.

#### Other notes
* `-DGAMMARAY_PROBE_ONLY_BUILD=1` builds are currently not tested, because they require a more complicated setup involving an existing build tree. However, I do plan to work on an update to the CI configuration that would test them, by using a second build job (not second step) which depends on the first, and uses the existing build trees from the standard build to perform a second probe-only build for a different Qt version. A future PR will add testing of probe-only build runs.

* The CI workflow file incorporates the `linuxdeployqt` steps from the old Travis CI configuration. However, they are currently disabled with an `if: false` rule. This is necessary due to the AppImage tools' shortsighted **requirement** that AppImages be generated on the oldest possible platforms available.

  That platform is currently `ubuntu-18.04`, which is both (a) deprecated by GitHub, and more importantly, (b) shipped with a glibc version that lacks the GLIBC_2.28 symbols required to build GammaRay's `master` branch. This makes building an AppImage of GammaRay impossible, based on the rules imposed by the AppImage project.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,71 @@
+name: CI Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        build-type:
+          - ''
+          - '-DGAMMARAY_CLIENT_ONLY_BUILD=1'
+#          - '-DGAMMARAY_PROBE_ONLY_BUILD=1 -DGAMMARAY_BUILD_UI=0'
+        qt-version: ['5.15.*', '6.4.*']
+
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


dantti:
Thanks, closing in favor of #838

PR #767, opened 2023-03-09
Fix CI by switching to JS markdownlint, adding override comment to README

ferdnyc:
The README currently fails markdownlint in the pre-commit CI, because the first line contains an HTML tag that pushes the line length over 80 columns.

@DavidAnson has a (node) JS implementation of markdownlint which offers a number of features over the previous Ruby-based implementation, including the ability to selectively disable/enable rules inline to the linted files via HTML comments.

Switching pre-commit to use DavidAnson/markdownlint-cli2, and adding a comment before the README's first line to disable the line-length check, should allow the file to pass the markdownlint check.

PR #769, opened 2023-03-11
Exclude markdownlint rule MD041

ferdnyc:
Fix the pre-commit CI checks, by ignoring MarkdownLint rule 41. (Replacement for #767, as requested by @winterz.)


ferdnyc:
My mistake, the issue has nothing to do with MD041.

PR #774, opened 2023-04-07
Try seeking the stream to the correct position on deserialization fai…

Waqar144:
…lure

If we get some data that can't be deserialized e.g., a User type that somehow made it to the client, all subsequent data fails to get deserialized and gammaray can crash.

To avoid this add a marker inside the data to restore reading from the stream in case we get something bad.

This is especially important with Qt6.4 because in Qt6 if a type defines QDatastream operator<< then our RemoteModel will happily send it across to the client.


client/remotemodel.cpp
```diff
@@ -390,13 +390,21 @@ void RemoteModel::newMessage(const GammaRay::Message &msg)
...
             typedef QHash<int, QVariant> ItemData;
             ItemData itemData;
             qint32 flags;
-            msg >> itemData >> flags;
+            // read item data
+            msg >> itemData;
+            // skip the marker and reset if itemData was invalid/unreadable
+            msg.findAndSkipCString(GammaRay::REMOTE_MODEL_MARKER, indexEndPos);
+            msg >> flags;
+            // msg >> itemData >> flags;
```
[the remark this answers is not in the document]
Waqar144: Done.

client/remotemodel.cpp
```diff
@@ -390,13 +390,20 @@ void RemoteModel::newMessage(const GammaRay::Message &msg)
                 qWarning() << "Unexpected empty index, probably some type failed to deserialize" << Q_FUNC_INFO;
                 continue;
             }
+            quint32 indexEndPos = 0;
+            msg >> indexEndPos;
```
dfaure-kdab, line 394: Is this streamed size (redundant with where it actually is in the stream) needed? Wouldn't `int indexEndPos = payload().device()->pos();` do the job maybe?

Waqar144: Indeed, thanks.

common/message.h
```diff
@@ -111,6 +111,26 @@ class GAMMARAY_COMMON_EXPORT Message
...
+    /**
+     * Finds @p marker and seeks the internal QDataStream to
+     * right after the marker.
+     *
+     * @p from the pos to start search from
+     */
+    void findAndSkipCString(const char *marker, int from) const;
+
+    /**
+     * Write a c string to the message of size == @p len
```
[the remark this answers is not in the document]
Waqar144: Done.

client/remotemodel.cpp
```diff
@@ -390,13 +390,23 @@ void RemoteModel::newMessage(const GammaRay::Message &msg)
                 qWarning() << "Unexpected empty index, probably some type failed to deserialize" << Q_FUNC_INFO;
                 continue;
             }
+            // in case of failure in reading itemData
+            // we will use this position to start looking
+            // for our special marker and restore the stream
+            // state
+            quint32 indexEndPos = msg.size();
```
dfaure-kdab, line 397: Seems wrong, this has to be "current pos", not "total size".

I think you need `Message::pos() { return payload().device()->pos(); }`  (and it could be used in message.cpp:269)

Waqar144: Yep, done now. Pushed too early.

dfaure-kdab (approved):

PR #780, opened 2023-04-21
Launcher/Attach: Focus on search/filter text field at startup

ratijas:
This allows immediately start typing filter query without hitting Tab key first.


Waqar144:
Thanks

PR #783, opened 2023-05-08
Fix invokeMethod failure with alias param

Waqar144:
Starting from Qt 6.5, invokeMethod has changed a lot. For normal users this doesn't matter too much, but for us, who still use the QGenericArgument way it breaks things. InvokeMethod with QGenericArg no longer checks aliases of a given argument and thus fails to call the method.

To work around this, use the actual types in the signals instead.

The QGenericArgument overload of invokeMethod is marked for removal in Qt7, so we might need to do it by hand in future but for now this will work.

### DIFF
--- a/plugins/quickinspector/quickinspectorinterface.cpp
+++ b/plugins/quickinspector/quickinspectorinterface.cpp
@@ -58,6 +58,7 @@ QuickInspectorInterface::QuickInspectorInterface(QObject *parent)
 {
     ObjectBroker::registerObject<QuickInspectorInterface *>(this);
     StreamOperators::registerOperators<Features>();
+    qRegisterMetaType<Features>("QFlags<GammaRay::QuickInspectorInterface::Feature>");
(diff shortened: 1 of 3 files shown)


plugins/quickinspector/quickinspectorinterface.cpp
```diff
@@ -58,6 +58,7 @@ QuickInspectorInterface::QuickInspectorInterface(QObject *parent)
 {
     ObjectBroker::registerObject<QuickInspectorInterface *>(this);
     StreamOperators::registerOperators<Features>();
+    qRegisterMetaType<Features>("QFlags<GammaRay::QuickInspectorInterface::Feature>");
```
vimpostor, line 61: Is this really needed? For me this works without this extra register call too.

Waqar144: This is needed for Qt5, the alias needs to be registered otherwise calls via invokeMethod will fail

PR #788, opened 2023-05-11
Use qhelpgenerator to build .qhc file

### LINKED ISSUE #569: The "qcollectiongenerator" tool is deprecated, use "qhelpgenerator" instead.  Failed to build with MSVC 2017 and Visual Studio 2019.
```
[100%] Built target gammaray-manual.qch                                                                                               
Scanning dependencies of target gammaray_qhc                                                                                          
[100%] Generating gammaray.qhc                                                                                                        
The "qcollectiongenerator" tool is deprecated, use "qhelpgenerator" instead.                                                          
                                                                                                                                      
Unknown input file type.                                                                                                              
                                                                                                                                      
                                                                                                                                      
Usage:                                                                                                                                
                                                                                                                                      
qhelpgenerator <file> [options]                                                                                                       
                                                                                                                                      
  -o <output-file>       Generates a Qt compressed help                                                                               
                         called <output-file> (*.qch) for the                                                                         
                         Qt help project <file> (*.qhp).                                                                              
                         Generates a Qt help collection                                                                               
                         called <output-file> (*.qhc) for the                                                                         
                         Qt help collection project <file> (*.qhcp).                                                                  
                         If this option is not specified                                                                              
                         a default name will be used                                                                                  
                         (*.qch for *.qhp and *.qhc for *.qhcp).                                                                      
  -c                     Checks whether all links in HTML files                                                                       
                         point to files in this help project.                                                                         
  -s                     Suppresses status messages.                                                                                  
  -v                     Displays the version of                                                                                      
                         qhelpgenerator.                                                                                              
                                                                                                                                      
NMAKE : fatal error U1077: 'D:\Qt\5.13.0\msvc2017_64\bin\qcollectiongenerator.exe' : return code '0x1'                                
Stop.                                                                                                                                 
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\bin\HostX64\x64\n
make.exe"' : return code '0x2'                                                                                                        
Stop.                                                                                                                                 
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\bin\HostX64\x64\n
make.exe"' : return code '0x2'                                                                                                        
Stop.                                                                                                                                 
```
Everything is visible in the log above.

ferdnyc:
In Qt5 and Qt6, `qhelpgenerator` can process `.qhcp` files as well as `.qhp` files. `qcollectiongenerator` is deprecated in Qt5 and removed in Qt6.

Fixes #569

### DIFF
--- a/docs/collection/CMakeLists.txt
+++ b/docs/collection/CMakeLists.txt
@@ -29,8 +29,7 @@ add_custom_command(
     OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/gammaray.qhc
     COMMAND ${CMAKE_COMMAND} -E copy ${manual_qch_file} ${CMAKE_CURRENT_BINARY_DIR}
     COMMAND ${CMAKE_COMMAND} -E copy ${apidocs_qch_file} ${CMAKE_CURRENT_BINARY_DIR}
-    COMMAND Qt::qhelpgenerator ${CMAKE_CURRENT_BINARY_DIR}/gammaray.qhcp -o
-            ${CMAKE_CURRENT_BINARY_DIR}/gammaray.qhc
+    COMMAND Qt::qhelpgenerator ${CMAKE_CURRENT_BINARY_DIR}/gammaray.qhcp -o ${CMAKE_CURRENT_BINARY_DIR}/gammaray.qhc
     DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/gammaray.qhcp.cmake ${manual_qch_target} ${apidocs_qch_target}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     COMMENT "Generate the GammaRay .qhc file using qhelpgenerator"
--- a/docs/collection/CMakeLists.txt
+++ b/docs/collection/CMakeLists.txt
@@ -6,6 +6,7 @@
 #
 # Contact KDAB at <info@kdab.com> for commercial licensing options.
 #
+find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS Help ToolsTools)
 if(NOT TARGET Qt::qhelpgenerator)
     message(STATUS "qhelpgenerator not found, documentation collection will not be generated.")
     return()
--- a/docs/collection/CMakeLists.txt
+++ b/docs/collection/CMakeLists.txt
@@ -6,8 +6,8 @@
 #
 # Contact KDAB at <info@kdab.com> for commercial licensing options.
 #
-if(NOT TARGET Qt5::qcollectiongenerator)
-    message(STATUS "qcollectiongenerator not found, documentation collection will not be generated.")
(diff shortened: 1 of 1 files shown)


docs/collection/CMakeLists.txt
```diff
@@ -6,8 +6,9 @@
 #
 # Contact KDAB at <info@kdab.com> for commercial licensing options.
 #
-if(NOT TARGET Qt5::qcollectiongenerator)
-    message(STATUS "qcollectiongenerator not found, documentation collection will not be generated.")
+find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS Help ToolsTools)
```
winterz, line 9: I think ToolsTools is not available for Qt5.
could you move `find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENT ToolsTools)` inside a conditional for !Qt5 or somesuch?

ferdnyc: @winterz Correct, it's not available for Qt5. And `AttributionsScannerTools` isn't available in Qt6. That's why these are all optional components.

This `find_package()` will import the CMake targets for `Qt::qdoc` and friends for either Qt5 or Qt6, even though those targets are defined in **different places** in the two versions. The ones that match the version in use will be loaded, the others will be unavailable.

winterz:
what I'm trying to solve is this bit of confusion from CMake (for Qt5 builds):

```
-- The following OPTIONAL packages have not been found:

 * Qt5IviCore (required version >= 1.1)
 * Qt5IviVehicleFunctions (required version >= 1.1)
 * Qt5IviMedia (required version >= 1.1)
 * Qt5ToolsTools
 ```

but there is no Qt5ToolsTools so CMake will always complain that its missing

docs/collection/CMakeLists.txt
ferdnyc: Anyway, yes, they can be split up into separate branches if you'd prefer, sure. There's no _harm_ in trying to import Qt5ToolsTools that will never exist, but it also isn't an imperative that we try to import it under Qt5.

Though, with that other branch I mentioned, I was actually planning to expand this with _additional_ (optional) imports to discover `qdoc` and `qattributionsscanner` using CMake targets as well (when possible), and some of those will not only not-exist in one of Qt5 or Qt6, but whether they exist will depend on the _version_ of Qt5 in use. Versions earlier than ~Qt 5.13 will be missing some CMake configs that _will_ exist in later versions, so there will still be potentially-nonexistent optional imports in the Qt5 branch. (The code would fall back to the current `find_program()`-based discovery, in those cases. But only as a fallback if the targets can't be imported from the optional components.)

However, `Qt5ToolsTools` will never exist under any circumstance, it's true.

ferdnyc:
\*nod\* Gotcha. I'll make it Qt6-only then, sure.

PR #794, opened 2023-05-19
Docs build: Prefer CMake targets for build tools

ferdnyc:
Instead of using `find_program()` to discover tools like `qdoc`, `qhelpgenerator`, and `qtattributionsscanner`, default to picking up those tools' locations from the relevant CMake IMPORTED executable targets.

Fall back to `find_program()` for any targets that don't exist, to preserve compatibility with Qt 5.12 and earlier.

I've successfully used the code on this branch to build GammaRay, complete with all documentation including the collections file, on:

* Ubuntu 20.04, using no options at all to configure for Qt 5.12:
  ```sh
  cmake -B _build -S .
  cmake --build _build
  ```

Because of the fallbacks, this build actually configures identically to the current `master` branch code.

* Fedora 38, setting only `GAMMARAY_QT6_BUILD` and `QDOC_INDEX_DIR` for Qt 6.5. (Because nothing can change the fact that Fedora doesn't ship a `qtcore.index` with their Qt6 documentation packages.)
  ```sh
  cmake -B _build -S . -DGAMMARAY_QT6_BUILD=1 -DQDOC_INDEX_DIR=/usr/share/doc/qt6
  cmake --build _build
  ```

Normally on Fedora 38, (using the `master` branch), at least `-DQT_INSTALL_BINS=/usr/lib64/qt6/libexec` would also have to be set, because `qtattributionsscanner` is not on the `$PATH`. But the target created by `Qt6ToolsTools` allows it to be referenced in CMake as `Qt::qtattributionsscanner`. 

Whereas for `qdoc`, (again on Fedora) `/usr/bin/qdoc` is the Qt5 version; the Qt6 version is named `/usr/bin/qdoc-qt6`. Using a `find_program()`-discovered `/usr/bin/qdoc` in a Qt6 build will cause it to fail. The `Qt::qdoc` CMake target, however, references the correct version of the tool for any Qt release in which it's available (which is Qt 5.13+).

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -8,28 +8,48 @@
 #
 
 # qdoc toolchain
-# TODO: some of this probably should be upstreamed to Qt's cmake files...
-# TODO replace with Qt::qhelpgenerator for Qt >= 5.7.1
-find_program(QHELPGEN_EXECUTABLE qhelpgenerator HINTS ${QT_INSTALL_BINS})
-find_program(QDOC_EXECUTABLE qdoc HINTS ${QT_INSTALL_BINS})
-find_program(QTATTRIBUTIONSSCANNER_EXECUTABLE qtattributionsscanner HINTS ${QT_INSTALL_BINS})
+find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS Help)
+if(Qt_VERSION_MAJOR EQUAL 5)
+    find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS AttributionsScannerTools)
+elseif(Qt_VERSION_MAJOR GREATER_EQUAL 6)
+    find_package(Qt${Qt_VERSION_MAJOR} NO_MODULE QUIET OPTIONAL_COMPONENTS ToolsTools)
+endif()
+
+if(TARGET Qt::qdoc)
+    set(QDOC_EXECUTABLE "$<TARGET_FILE:Qt::qdoc>")
+else()
+    find_program(QDOC_EXECUTABLE qdoc HINTS ${QT_INSTALL_BINS})
+endif()
+
+if(TARGET Qt::qhelpgenerator)
+    # Required for Doxygen
+    get_target_property(QHELPGEN_EXECUTABLE Qt::qhelpgenerator IMPORTED_LOCATION)
+else()
+    find_program(QHELPGEN_EXECUTABLE qhelpgenerator HINTS ${QT_INSTALL_BINS})
+endif()
+
+if(TARGET Qt::qtattributionsscanner)
+    set(QTATTRIBUTIONSSCANNER_EXECUTABLE Qt::qtattributionsscanner)
+else()
+    find_program(QTATTRIBUTIONSSCANNER_EXECUTABLE qtattributionsscanner HINTS ${QT_INSTALL_BINS})
+endif()
 
 if(NOT DEFINED QDOC_TEMPLATE_DIR)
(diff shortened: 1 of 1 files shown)


docs/CMakeLists.txt
```diff
@@ -8,28 +8,48 @@
...
+    get_target_property(QHELPGEN_EXECUTABLE Qt::qhelpgenerator IMPORTED_LOCATION)
+else()
+    find_program(QHELPGEN_EXECUTABLE qhelpgenerator HINTS ${QT_INSTALL_BINS})
+endif()
+
+if (TARGET Qt::qtattributionsscanner)
+    set(QTATTRIBUTIONSSCANNER_EXECUTABLE Qt::qtattributionsscanner)
+else()
+    find_program(QTATTRIBUTIONSSCANNER_EXECUTABLE qtattributionsscanner HINTS ${QT_INSTALL_BINS})
+endif()
```
ferdnyc: Three different tools, three different ways of consuming the target. Because:

1. `qdoc` is used in `add_custom_command()` as an argument to `${CMAKE_COMMAND} -E env ...` commands, meaning that the actual path to the binary has to be supplied. `$<TARGET_FILE:Qt::qdoc>` is the recommended method of doing so, according to [the CMake docs for `add_custom_command`]()>
   > Arguments to COMMAND may use [generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#manual:cmake-generator-expressions(7)). Use the [TARGET_FILE](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_FILE) generator expression to refer to the location of a target later in the command line (i.e. as a command argument rather than as the command to execute).

2. `Qt::qtattributionsscanner` is used directly as the `COMMAND` supplied to `add_custom_command()`, which means that the target name can be used as the command.

3. The `Qt::qhelpgenerator` binary path needs to be supplied directly in the `QHELPGEN_EXECUTABLE` variable, as it'll be substituted into the generated Doxyfile for the API doc generation.

docs/CMakeLists.txt
```diff
@@ -40,12 +60,25 @@ if(NOT QDOC_EXECUTABLE
...
+    endforeach()
+
     message(STATUS "Unable to build user manual in qch format.")
     message(STATUS "qdoc executable: ${QDOC_EXECUTABLE}")
     message(STATUS "qhelpgenerator executable: ${QHELPGEN_EXECUTABLE}")
     message(STATUS "qtattributionsscanner executable: ${QTATTRIBUTIONSSCANNER_EXECUTABLE}")
-    message(STATUS "qdoc template dir: ${QDOC_TEMPLATE_DIR}")
-    message(STATUS "qdoc index dir: ${QDOC_INDEX_DIR}")
+    message(STATUS "qdoc template: ${QDOC_TEMPLATE}")
+    message(STATUS "qdoc index: ${QDOC_INDEX}")
```
ferdnyc, line 81: I slightly went back on my previous change, here, and switched back to displaying the `QDOC_TEMPLATE` and `QDOC_INDEX` (renamed from `QDOC_QTCORE_INDEX`) variables rather than the directory variables. That's so that, if the `find_program()` command is used and fails to find a suitable binary, `QDOC_TEMPLATE-NOTFOUND` or `QDOC_INDEX-NOTFOUND` will be displayed as the value, rather than an empty string.

The overly-complicated bit of code following "# Expand imported target data for status reporting", above, ensures that the variables being reported have sensible values, in the cases where targets are used instead of fallback `find_program()` calls.

ferdnyc: All of this only ever runs in cases where some part of the help-tool-discovery logic _fails_, so hopefully most of the time it won't even be triggered.

PR #796, opened 2023-05-22
Choose Qt version based on GAMMARAY_QT6_BUILD only

dantti:


CMakeLists.txt
```diff
@@ -414,29 +414,26 @@ if(GAMMARAY_ENFORCE_QT_ASSERTS)
     add_definitions(-DQT_FORCE_ASSERTS)
 endif()
 
+gammaray_option(GAMMARAY_QT6_BUILD "Build with Qt6." OFF)
+if(GAMMARAY_QT6_BUILD)
+    set(Qt_VERSION_MAJOR 6)
+    set(QT_MIN_VERSION "6.0.0")
+    set(QT_MAJOR_VERSION 6) #needed by ECM cmake
+else()
+    set(QT_MAJOR_VERSION 5) #needed by ECM cmake
```
redstrate, line 423: What is `Qt_VERSION_MAJOR` defined as when `GAMMARAY_QT6_BUILD` is OFF? I suspect it's just nothing, but I would define it just to make sure it's 5. (Or even better, just grep & replace all instances of `Qt_VERSION_MAJOR` with `QT_MAJOR_VERSION`)

PR #797, opened 2023-05-22
Use QT_VERSION_MAJOR in order to choose which Qt version to use

dantti:
Defining QT_VERSION_MAJOR will force the desired
Qt version as well as selecting the right "kit" in QtCreator, if not defined it will follow what
CMAKE_PREFIX_PATH provides first.

This drop the currently not working
GAMMARAY_QT6_BUILD option.


Waqar144 (approved):

Waqar144:
cmake presets need adjustment

PR #809, opened 2023-05-26
Micro optimizations for ObjectTreeModel

Waqar144:


milianw:
lgtm, but does the inlining of addressToString actually improve anything?

nevertheless, +2 from my side

Waqar144:
I have seen some improvement but nothing concrete, but I think it will be hard to measure the cost of inlining when the app is run as a whole. The reason for inlining it was that the function is small, will not affect compiles as nothing needs to be added to the header.

Thanks for reviewing

PR #832, opened 2023-09-07
Fix non-system include in kuserfeedback

ferdnyc:
The include of local header
common/surveytargetexpressionevaluator.h
isn't a system include, in context.

### DIFF
--- a/3rdparty/kuserfeedback/core/provider_p.h
+++ b/3rdparty/kuserfeedback/core/provider_p.h
@@ -9,7 +9,7 @@
 
 #include "provider.h"
 
-#include <common/surveytargetexpressionevaluator.h>
+#include "common/surveytargetexpressionevaluator.h"
 
 #include <QDateTime>
 #include <QStringList>


ferdnyc:
Note2: I've also submitted this change upstream:
https://invent.kde.org/libraries/kuserfeedback/-/merge_requests/46

Waqar144:
Was it causing any issues? If not, then I'd prefer to leave it as is and we'll get this change automatically once upstream has it and we rebase on it on future.

ferdnyc:
@Waqar144 

> Was it causing any issues? 

Apparently not, not even a warning. I noticed the patch in my PR to update Fedora's packaging, and noticed it hadn't been upstreamed... but I just removed it, and it seemingly has no effect whatsoever. So I'll just drop it from that end, too.

Sorry about that, no idea why it's even there. Looks like it dates all the way back to March, whatever I was thinking at the time is long forgotten.

PR #838, opened 2023-09-22
Add GitHub actions

dantti:


winterz:
@dantti can you somehow fold this together with https://github.com/KDAB/GammaRay/pull/768

dantti:
@winterz so PR https://github.com/KDAB/GammaRay/pull/768 did not enable tests, and had a disabled AppImage step, that uses the old linuxdeploy (not the one we currently use for Qt6 stuff), also I don't think GitHub actions should create Gammaray binaries because gammaray is quite tied to the same Qt version the app is running.

A full run takes quite some time so that perhaps we would only want github actions to build packages for tagged versions.

This PR also enables all tests that passes, I added a comment on each configuration that I skip tests, as they depend on Qt version and OS.

IMO it's good to get this merged now, and we can fix tests as we manage to, will we retire buildbot for this one?

.github/workflows/build.yml
```diff
@@ -0,0 +1,163 @@
...
+          #- Release
+        client_only:
+          - client_only
+          - client_and_ui
+        qt:
+          - version: "5.15.2"
+            requested: "5.15"
+          - version: "6.5.3"
+            requested: "6.5.*"
+            modules: qtshadertools
```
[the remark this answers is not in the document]
ferdnyc: Come to think of it, I don't think that will quite work for Windows PowerShell, but this will:

```yaml
- run: >
    ${{ runner.os == 'windows'
    && '"QT_VERSION=$(qtpaths --qt-version)" >> $env:GITHUB_ENV' }}
    || 'echo "QT_VERSION=$(qtpaths --qt-version)" >> "$GITHUB_ENV"' }}

File: .github/workflows/build.yml [hunk 2]
```diff
@@ -0,0 +1,163 @@
...
+        # Exclude
+        # quicktexturetest
+        # bindinginspectortest
+        # modelinspectortest fails in Qt6/CI passes locally
+      - name: Qt6 Run tests on Windows
+        if: ${{ matrix.build_type == 'Debug' && runner.os == 'Windows' && startsWith(matrix.qt.version, '6.') }}
+        run: >
+          ctest --test-dir ./build -C ${{ matrix.build_type }} --output-on-failure
+          --exclude-regex "quicktexturetest|launchertest|bindinginspectortest|modelinspectortest"
+
```
ferdnyc, line 158: This will _work_, no question. But except for the exclusion list, the test instances are basically identical. (99%, anyway — just the Linux envvars.)

IMHO it's cleaner to just use one definition, and break out the selection of exclusion list into a previous step. I got it to work like this, which is a little clunky but pretty readable...

```yaml
jobs:
  build:
# [...]
    env:
      Windows_5: 'quicktexturetest|bindinginspectortest'
      Windows_6: 'quicktexturetest|launchertest|bindinginspectortest|modelinspectortest'
      macOS_5: 'connectiontest-*|probeabidetectortest|launchertest|quickinspectortest2'
      macOS_6: 'connectiontest-*|probeabidetectortest|launchertest|clientconnectiontest|modelinspectortest|quickinspectortest2' 
      Linux_5: 'quickmaterialtest|quicktexturetest|connectiontest-style-filter'
      Linux_6: 'quickmaterialtest|quicktexturetest|connectiontest-style-filter|bindinginspectortest|quickinspectortest|quickinspectortest2|modelinspectortest'
      QT_QPA_PLATFORM: ${{ contains('ubuntu', matrix.os) && 'offscreen' || '' }}
      QT_QUICK_BACKEND: ${{ contains('ubuntu', matrix.os) && 'software' || '' }}
    steps:
      - name: "Select exclusions"
        env:
          CTEST_EXCLUSIONS: >
            ${{ (runner.os == 'Windows' && matrix.qt.version == 5) && env.Windows_5 || '' }}
            ${{ (runner.os == 'Windows' && matrix.qt.version == 6) && env.Windows_6 || '' }}
            ${{ (runner.os == 'Linux' && matrix.qt.version == 5) && env.Linux_5 || '' }}
            ${{ (runner.os == 'Linux' && matrix.qt.version == 6) && env.Linux_6 || '' }}
            ${{ (runner.os == 'macOS' && matrix.qt.version == 5) && env.macOS_5 || '' }}
            ${{ (runner.os == 'macOS' && matrix.qt.version == 6) && env.macOS_6 || '' }}
        run: >
          ${{ runner.os == 'windows'
          && '"CTEST_EXCLUSIONS=$env:CTEST_EXCLUSIONS" >> $env:GITHUB_ENV'
          || 'echo "CTEST_EXCLUSIONS=$CTEST_EXCLUSIONS" >> "$GITHUB_ENV"' }}
```

It defines all of the right envvars for each system, so they can be dropped in later (Including `--exclude-regex "${{ env.CTEST_EXCLUSIONS }}"` for the CTest invocation.)

Though now I'm also seeing that I have to restrict the tests to _non_-client-only builds. A bunch of them complain if they're missing a probe for the platform.

ferdnyc: (The above also relies on a `matrix.qt` setup of the form I suggested earlier:)

```yaml
matrix:
  qt:
    - version: 5
      request: 5.15.*
    - version: 6
      request: 6.4.*
      modules: qtshadertools
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,163 @@
...
+          modules: ${{ matrix.qt.modules }}
+          version: ${{ matrix.qt.requested }}
+          cache: true
+
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Create build directory
+        run: mkdir build
+
```
ferdnyc, line 52: It's really unnecessary to waste a step on this, CMake does it for you.

.github/workflows/build.yml
```diff
@@ -0,0 +1,163 @@
...
+            requested: "5.15"
+          - version: "6.5.3"
+            requested: "6.5.*"
+            modules: qtshadertools
+
+    steps:
+      - name: Install Qt with options and default aqtversion
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: null # use whatever the default is
```
ferdnyc, line 42: I'm trying to understand why this is here.

If it's not specified _at all_, the default will be used, right? That's why it's the default. Why would it be necessary to explicitly **ask** for a default?

ferdnyc (commented):
A few notes.

Regarding #768, the only reason I included things like the AppImage build (disabled, as it can't work, for reasons I explained in the comments/PR) was because I figured resurrecting the old Travis CI script was as good a starting point as any.

I'd have happily removed that, or made any other changes if requested. But the PR's been there since March without any indication that it'd even been noticed. A comment on _this_ PR is the **very first time** it's ever been acknowledged or mentioned.

.github/workflows/build.yml
```diff
@@ -0,0 +1,163 @@
...
+
+      - name: Make sure MSVC is found when Ninja generator is in use
+        if: ${{ runner.os == 'Windows' }}
+        uses: ilammy/msvc-dev-cmd@v1
+
+      - name: Configure project
+        run: >
+          cmake -S . -B ./build -G Ninja
+          -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+          -DGAMMARAY_CLIENT_ONLY_BUILD=${{ matrix.client_only.client_only }}
```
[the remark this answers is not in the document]
ferdnyc: Oh ­— actually, this _doesn't_ work. All of the configured jobs have `-DGAMMARAY_CLIENT_ONLY_BUILD=` on the CMake command line, which is why your tests didn't fail due to the missing probes that it _would_ otherwise have caused.

(Also, Windows client only builds are currently broken due to some more bad target-creation logic I didn't catch (because it doesn't affect Linux) in `plugins/quickinspector/CMakeLists.txt`. I'll have a PR submitted in a couple of minutes to fix that.)

.github/workflows/build.yml
ferdnyc: (If you don't want to pollute the process environment, the exclusions could also be baked directly into the selection step, e.g.:)

```yaml
# [...]
    env:
      QT_QPA_PLATFORM: ${{ contains('ubuntu', matrix.os) && 'offscreen' || '' }}
      QT_QUICK_BACKEND: ${{ contains('ubuntu', matrix.os) && 'software' || '' }}
    steps:
      - name: "Select exclusions"
        env:
          CTEST_EXCLUSIONS: >
            ${{ (runner.os == 'Windows' && matrix.qt.version == 5)
            && 'quicktexturetest|bindinginspectortest' || '' }}
            ${{ (runner.os == 'Windows' && matrix.qt.version == 6) 
            && 'quicktexturetest|launchertest|bindinginspectortest|modelinspectortest' || '' }}
            ${{ (runner.os == 'Linux' && matrix.qt.version == 5) 
            && 'quickmaterialtest|quicktexturetest|connectiontest-style-filter' || '' }}
            ${{ (runner.os == 'Linux' && matrix.qt.version == 6) 
            && 'quickmaterialtest|quicktexturetest|connectiontest-style-filter|bindinginspectortest|quickinspectortest|quickinspectortest2|modelinspectortest' || '' }}
            ${{ (runner.os == 'macOS' && matrix.qt.version == 5) 
            && 'connectiontest-*|probeabidetectortest|launchertest|quickinspectortest2' || '' }}
            ${{ (runner.os == 'macOS' && matrix.qt.version == 6) 
            && 'connectiontest-*|probeabidetectortest|launchertest|clientconnectiontest|modelinspectortest|quickinspectortest2' || '' }}
        run: >
          ${{ runner.os == 'windows'
          && '"CTEST_EXCLUSIONS=$env:CTEST_EXCLUSIONS" >> $env:GITHUB_ENV'
          || 'echo "CTEST_EXCLUSIONS=$CTEST_EXCLUSIONS" >> "$GITHUB_ENV"' }}
```

That'll have the same effect, without defining all the `Linux_5`, `Windows_6`, etc. envvars that'd otherwise persist through the entire job run. The comments from the individual steps in the current workflow could also, **I _believe_**, be included between each condition without affecting the final result.

.github/workflows/build.yml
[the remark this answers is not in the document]
dantti: AppImage can be created just fine by using qt plugin of Linux deploy, still it's too slow for regular builds.

Qt version is perfectly fine this way, as we might want to have more versions to test regressions or behavior changes. The version text is just text, what is used is the requested one, this is from the Qt git hub actions  page, which I might change in future to look a bit more like KDDockWidget, but there I create a much more complex setup so for now this is enough.

.github/workflows/build.yml
```diff
@@ -0,0 +1,163 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false #true
```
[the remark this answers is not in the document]
dantti: Was left there to remember to return to true, and it worked ;)

.github/workflows/build.yml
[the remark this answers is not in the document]
dantti: right, I initially had true and false but was not readable on the github matrix, do forgot to update it

.github/workflows/build.yml
dantti: We can remove ofcourse, but I feel like this might be needed in future, also helps one knowing that it uses aqt behind the scenes

.github/workflows/build.yml
dantti: Sorry but IMO it's a lot easier to have multiple steps that are way easier to read.
This workaround GitHub actions lack of ternary options is not obvious so I prefer to avoid it.

.github/workflows/build.yml
ferdnyc: That's fair.

ferdnyc:
So, some of my review comments on the exclusion structure aren't entirely accurate.

I _did_ eventually get it working, like this:

```yaml
env:
  Windows_5: 'quicktexturetest|bindinginspectortest'
  Windows_6: 'quicktexturetest|launchertest|bindinginspectortest|modelinspectortest'
  macOS_5: 'connectiontest-*|probeabidetectortest|launchertest|quickinspectortest2'
  macOS_6: 'connectiontest-*|probeabidetectortest|launchertest|clientconnectiontest|bindinginspectortest|modelinspectortest|quickinspectortest2' 
  Linux_5: 'quickmaterialtest|quicktexturetest|connectiontest-style-filter'
  Linux_6: 'quickmaterialtest|quicktexturetest|connectiontest-style-filter|bindinginspectortest|quickinspectortest|quickinspectortest2|modelinspectortest'

jobs:
  build:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
        build_type: [Debug]
        client_only: [ON, OFF] 
        qt:
          - { version: 5, request: 5.15.*, modules: }
          - { version: 6, request: 6.5.*, modules: qtshadertools }
    runs-on: ${{ matrix.os }}
    steps:
      - name: "Select exclusions"
        env:
          CTEST_EXCLUSIONS: ${{ (runner.os == 'Windows' && matrix.qt.version == 5) && env.Windows_5 || '' }}${{ (runner.os == 'Linux' && matrix.qt.version == 5) && env.Linux_5 || '' }}${{ (runner.os == 'Linux' && matrix.qt.version == 6) && env.Linux_6 || '' }}${{ (runner.os == 'macOS' && matrix.qt.version == 5) && env.macOS_5 || '' }}${{ (runner.os == 'macOS' && matrix.qt.version == 6) && env.macOS_6 || '' }}
        run: >
          ${{ runner.os == 'windows'
          && '"CTEST_EXCLUSIONS=$env:CTEST_EXCLUSIONS" | Out-File -FilePath $env:GITHUB_ENV -Append'
          || 'echo "CTEST_EXCLUSIONS=$CTEST_EXCLUSIONS" >> "$GITHUB_ENV"' }}

      - name: "Configure Linux test environment"
        if: runner.os == 'Linux' && matrix.client_only == 'OFF'
        run: |
          echo "QT_QPA_PLATFORM=offscreen" >> "$GITHUB_ENV"
          echo "QT_QUICK_BACKEND=software" >> "$GITHUB_ENV"
```

The envvar definition has to be all on one line like that, or the value ends up with spaces in it that throw off the expression matching. (GitHub really needs to add a `trim()` function to their `${{ }}` expression syntax.)

Which admittedly makes it kind of ugly. Accusations that I'm being too pointlessly "clever" graciously accepted.

I should've mentioned, BTW: the latest (still in-progress, as of this moment) run of my version of the workflow can be seen here: https://github.com/ferdnyc/GammaRay/actions/runs/6397369854

I had to add the `bindinginspectortest` to the macOS 6 exclusions, as well, and I'm still waiting to see what the final outcome of the Windows attempts will be.

The full workflow itself is: https://github.com/ferdnyc/GammaRay/blob/master/.github/workflows/test-trickery.yml

dantti:
Well IMO this huge one line is quite unreadable, plus, this should be temporary, we need these tests either permanently skipped or fixed

PR #841, opened 2023-10-03
Fix client-only builds with the quckinspector enabled

ferdnyc:
In the spirit of my previous #765, this fixes some places in the `plugins/quickinspector/CMakeLists.txt` where targets that weren't created during a `GAMMARAY_CLIENT_ONLY_BUILD` were being assigned to, causing CMake to abort. I missed these previously because they only affect Windows, as far as I can tell.

I'd suggest reviewing the PR changes as separate commits. Taken all together, it looks like a LOT, because the indentation of nearly every line in the file changed due to some `if()` nesting I removed. (Preferring to do an early return from the file, if we don't have the `Qt::Quick` target, rather than having the **entire** thing wrapped in `if(TARGET Qt::Quick)`.)

I also fixed a lot of formatting, chopping down long commands so that they're more vertical.

But the **only** three actual _code_ changes in the PR are:

1. Replace the outermost `if(TARGET Qt::Quick)` nesting with:
   ```cmake
   if(NOT TARGET Qt::Quick)
       return()
   endif()
   ```

2. Move the `endif()` for `if(NOT GAMMARAY_CLIENT_ONLY_BUILD)` farther down, so it protects more code that can't be run in client-only builds.

3. Replace this call:
   ```cmake
   set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} Qt::Gui_GLESv2)
   ```
   With:
   ```cmake
   list(APPEND CMAKE_REQUIRED_LIBRARIES Qt::Gui_GLESv2)
   ```

### DIFF
--- a/plugins/quickinspector/CMakeLists.txt
+++ b/plugins/quickinspector/CMakeLists.txt
@@ -68,21 +68,23 @@ if(NOT GAMMARAY_CLIENT_ONLY_BUILD)
         Qt::QuickPrivate
         gammaray_kitemmodels
     )
-endif()
 
-if(TARGET Qt::Gui_GLESv2)
-    include(CheckCXXSourceCompiles)
-    set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} Qt::Gui_GLESv2)
-    check_cxx_source_compiles("int main() {return 0;}" CAN_LINK_GUI_GLESV2)
-    if(CAN_LINK_GUI_GLESV2)
-        set(HAVE_GLESv2 TRUE)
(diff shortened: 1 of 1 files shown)


ferdnyc:
> I also fixed a lot of formatting, chopping down long commands so that they're more vertical.

Aaand, of course, the pre-commit hooks _**undid**_ all of that effort, and made the code look awful again.

The `cmake-format` settings here are... bad. They are not good.

winterz:
I don't understand why cla-assistant is asking about the CLA again.  I clearly see ferdnyc has already signed.

thanks for the patch. I'll commit.
yes, that's strange cmake formating.  maybe one day I'll take a closer look.

PR #853, opened 2023-11-08
Repair lack of classnames in "Graphics Scenes" or "Styles" combobox

dfaure-kdab:
Since Qt 6.3, QAbstractProxyModel::itemData() no longer calls data() but calls into the source model. So we need to reimplement it in order to use our data() reimplementation.

Should I target master or something else?


Waqar144 (approved):

Waqar144:
Needs be picked to 3.0 branch

PR #865, opened 2023-11-10
Show classnames next to objectNames, when different

dfaure-kdab:
Property     | Value                        | Type     | Class
Before:
  parent       | m_loadingWidget              | QObject* | QObject
After:
  parent       | m_loadingWidget (QSvgWidget) | QObject* | QObject

Should this be QSvgWidget* with a star?
Or something more consistent with the case where there's no objectName, which shows "QFrame[this=0xdcd22e0]" ?

The change propagated to other classes to fix const [in]correctness (the const_cast is only because of QHash API, it shouldn't be necessary for the caller of those methods).

### DIFF
--- a/core/objectdataprovider.cpp
+++ b/core/objectdataprovider.cpp
@@ -48,7 +48,7 @@ QString ObjectDataProvider::name(const QObject *obj)
     return name;
 }
 
-QString ObjectDataProvider::typeName(QObject *obj)
+QString ObjectDataProvider::typeName(const QObject *obj)
 {
     if (!obj)
         return QString();
@@ -61,7 +61,7 @@ QString ObjectDataProvider::typeName(QObject *obj)
     return obj->metaObject()->className();
 }
 
-QString ObjectDataProvider::shortTypeName(QObject *obj)
+QString ObjectDataProvider::shortTypeName(const QObject *obj)
 {
     if (!obj)
         return QString();
@@ -74,7 +74,7 @@ QString ObjectDataProvider::shortTypeName(QObject *obj)
     return obj->metaObject()->className();
 }
 
-SourceLocation ObjectDataProvider::creationLocation(QObject *obj)
+SourceLocation ObjectDataProvider::creationLocation(const QObject *obj)
 {
     SourceLocation loc;
     if (!obj)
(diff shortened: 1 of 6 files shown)


Waqar144:
> Should this be QSvgWidget* with a star?

I would prefer this as its slightly more accurate.

For the case where we don't have objectName, I think show `QFrame[this=0xdcd22e0]` is a good idea and is consistent with some other places.

dfaure-kdab:
`QFrame[this=0xdcd22e0]` is already implemented, I was only mentioning it because there's some inconsistency between `objectName(className)` and `className[this=value]`
The className is in parenthesis in one, outside everything in the other, and the pointer value is only shown in the case without an object name. So I was wondering if we should unify this somehow....
Maybe it could be className[objectName, this=value] ? I'm not 100% sure, because the object name might be the main thing we want to see.

Brainstorming other solutions:   `[this=value](className)` and `objectName[this=value](className)`?

About `QSvgWidget *` I'm also hesitant. In my example above, it's indeed a property that is a raw pointer. But we're in the code that prints information about an object, we don't know what the property in the parent object looks like. Maybe it's a shared_ptr. Maybe it's a unique_ptr. Maybe it's a reference (?). In such cases showing a raw pointer might be confusing?

Waqar144:
> ... Maybe it's a reference (?). In such cases showing a raw pointer might be confusing?

Right, I thought you were asking about whether to show an asterisk for the raw pointer case. Otherwise yeah, then just showing the className is enough. It also consumes less space (which is a somewhat important thing in this context)

Now that I saw the code and tried it out, I think I understand question a bit better. In the properties tab, just showing the objectName is more confusing imo. So I think `className[objectName, this=value]` and `className[this=ADDRESS]` is a good choice and consistent. The only downside I can see is the former taking more horizontal space, perhaps `this=value` can be skipped when we have object name.

ferdnyc:
> The className is in parenthesis in one, outside everything in the other, and the pointer value is only shown in the case without an object name. So I was wondering if we should unify this somehow....
> Maybe it could be `className[objectName, this=value]` ? I'm not 100% sure, because the object name might be the main thing we want to see.

I actually don't think standardizing around `{objectName_or_value}(ClassName)` would look terrible. e.g.

* `main(QMainWindow)`
* `toolDock(QDockWidget)`
* `0xdcd22e0(QFrame)`

(...Might encourage people to name their objects, too.)

Waqar144:
> I actually don't think standardizing around {objectName_or_value}(ClassName) would look terrible. e.g.

+1 from me too

Waqar144 (approved):

PR #881, opened 2023-12-12
Require Qt5.15 or above

winterz:
i.e no longer support Qt5 versions < 5.15

Future patches will remove code specific for Qt versions less than 5.15.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,7 +9,8 @@ To build GammaRay you will need *at least*:
 
 - CMake 3.16.0
 - a C++ compiler with C++11 support
-- Qt 5.5 or higher, or Qt 6
+- Qt 5.15 or higher, or Qt 6
+  (GammaRay v3.0.1 was the last release to support Qt versions less than 5.15)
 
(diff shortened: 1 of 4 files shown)


Waqar144 (approved):

INSTALL.md
```diff
@@ -9,7 +9,8 @@ To build GammaRay you will need *at least*:
 
 - CMake 3.16.0
 - a C++ compiler with C++11 support
-- Qt 5.5 or higher, or Qt 6
+- Qt 5.15 or higher, or Qt 6
+  (GammaRay v3.0.1 was the last release to support Qt versions less than 5.15)
```
ferdnyc, line 13: This seems incorrect, given 3.0.1's unreleased status. Since support for Qt < 5.15 was removed before the release of 3.0.1, then _3.0.0_ was the last release to support those older versions.

winterz: good point.
I fixed in 0012a5ab6a18a094a961dcb6551246b9da8ff6ba

PR #883, opened 2023-12-15
Remove conditional code supporting old Qt5 versions

winterz:
Remove support for Qt5 versions less than 5.15.


INSTALL.md
```diff
@@ -46,7 +46,7 @@ CMake option `CMAKE_PREFIX_PATH` and point it to the folder of your installation
 For example:
 
 ```bash
-    cmake -DCMAKE_PREFIX_PATH=$HOME/Qt/5.11.2/gcc_64 ..
+    cmake -DCMAKE_PREFIX_PATH=$HOME/Qt/5.15.4/gcc_64 ..
 ```
```
ferdnyc, line 50: Maybe this should just use `$HOME/Qt/6.6.0/gcc_64` as an example, instead? (Along with the envvar version below.)

Doesn't hurt to _encourage_ users to migrate to Qt6, after all.

ferdnyc (commented):
Wow, I had _no **idea**_ there was so MUCH legacy code in there. Hooray for streamlining! :+1:

common/recursiveproxymodelbase.h
```diff
@@ -22,28 +22,19 @@
...
 
-#if QT_VERSION < QT_VERSION_CHECK(5, 10, 0)
-#include <kde/krecursivefilterproxymodel.h>
-#define GAMMARAY_PROXY_BASE_CLASS KRecursiveFilterProxyModel
-#else
 #include <QSortFilterProxyModel>
 #define GAMMARAY_PROXY_BASE_CLASS QSortFilterProxyModel
-#endif
 
 class GAMMARAY_COMMON_EXPORT RecursiveProxyModelBase : public GAMMARAY_PROXY_BASE_CLASS
```
Waqar144, line 28: `GAMMARAY_PROXY_BASE_CLASS` -> `QSortFilterProxyModel`

common/recursiveproxymodelbase.h
```diff
@@ -22,28 +22,19 @@
...
-#if QT_VERSION < QT_VERSION_CHECK(5, 10, 0)
-    using KRecursiveFilterProxyModel::KRecursiveFilterProxyModel;
-#else
     RecursiveProxyModelBase(QObject *parent)
         : QSortFilterProxyModel(parent)
     {
         setRecursiveFilteringEnabled(true);
     }
 
     virtual bool acceptRow(int sourceRow, const QModelIndex &sourceParent) const;
```
Waqar144, line 37: This function can be removed as well in favor of the one below it, but we can do this as a followup

Waqar144 (changes requested):

INSTALL.md
winterz: good idea. I'll do that

common/recursiveproxymodelbase.h
winterz: will do

winterz: done

common/recursiveproxymodelbase.h
winterz: agreed

Waqar144 (approved):

PR #892, opened 2023-12-25
Disable statemachine example if missing deps

ferdnyc:
Just like commit 8921c0a did for the state machine viewer, the state machine example build should be skipped unless the necessary dependencies are available.


ferdnyc:
Yeah, I don't _think_ that CI failure has anything to do with this change. (Not really sure what happened there.)

PR #895, opened 2023-12-26
FindQmlLint: Prefer to use the Qt::qmllint target

### LINKED ISSUE #894: New qml_lint picks up Qt 5 version, fails the build
```
# kdesrc-build running: 'ninja' '-j' '6'
# from directory: /home/ratijas/kde/build6/gammaray
[1/24] Run qmlint on the specified files and create the associated timestamp semaphore
FAILED: examples/quick-batching/Slider.qml.qmllint /home/ratijas/kde/build6/gammaray/examples/quick-batching/Slider.qml.qmllint 
cd /home/ratijas/kde/build6/gammaray/examples/quick-batching && /usr/bin/qmllint /home/ratijas/kde/src6/gammaray/examples/quick-batching/Slider.qml && /usr/bin/cmake -E touch /home/ratijas/kde/build6/gammaray/examples/quick-batching/Slider.qml.qmllint
/home/ratijas/kde/src6/gammaray/examples/quick-batching/Slider.qml:16 : Expected token `;'
[2/24] Run qmlint on the specified files and create the associated timestamp semaphore
[3/24] Linking CXX shared module lib/gammaray/3.0/qt6_6-x86_64/gammaray_kjobtracker_plugin.so
[4/24] Building CXX object ui/CMakeFiles/gammaray_ui.dir/aboutdata.cpp.o
[5/24] Building CXX object translations/CMakeFiles/ljsonupdate.dir/ljsonupdate.cpp.o
[6/24] Building CXX object launcher/cli/CMakeFiles/gammaray.dir/main.cpp.o
ninja: build stopped: subcommand failed.
```

Clearly, /usr/bin/qmllint is coming from qt5-declarative package.

Besides, what the heck is this linter suggesting? It is really suggesting no whitespace between the => arrow function?

```js
            onClicked: (mouse)=> {
                updateCurrentIndex(mouse.x, mouse.y)
            }
```

ferdnyc:
If a target exists `Qt${QT_MAJOR_VERSION}::qmllint`, we can simply use that as our qmllint executable instead of searching around the path for one.

We have to keep the path-based discovery for Qt5, which doesn't have a target.

I also added a `qmllint_all`, to be the parent target of all of the linting targets. (Which are themselves backed by `add_custom_command(OUTPUT..` targets. It's targets all the way down.

Fixes #894

### DIFF
--- a/cmake/FindQmlLint.cmake
+++ b/cmake/FindQmlLint.cmake
@@ -8,49 +8,65 @@
 # Contact KDAB at <info@kdab.com> for commercial licensing options.
 #
 
-find_program(QMLLINT_EXECUTABLE qmllint)
-if(QMLLINT_EXECUTABLE)
-    if(NOT QMLLINT_IS_WORKING)
-        # Try to fix common problems on Debian-based distros -- they provide /usr/bin/qmllint,
-        # which is a symlink to /usr/lib/x86_64-linux-gnu/qt4/bin/qmllint (or the Qt5 version of it).
-        # The actual executable is part of different package, so might not even be installed =>
-        # double-check whether qmllint is working by executing it
-        execute_process(
-            COMMAND ${QMLLINT_EXECUTABLE} --version
-            RESULT_VARIABLE _qmllint_result
-            OUTPUT_QUIET ERROR_QUIET
-        )
-        if(_qmllint_result EQUAL 0)
-            set(QMLLINT_IS_WORKING
-                TRUE
-                CACHE BOOL "Whether the found qmllint executable is actually usable" FORCE
+set(QmlLint_EXECUTABLE)
+set(QmlLint_FOUND FALSE)
+
+# First check for a target (Qt6)
+if(TARGET Qt${QT_VERSION_MAJOR}::qmllint)
+    set(QmlLint_FOUND TRUE)
+    set(QmlLint_EXECUTABLE Qt${QT_VERSION_MAJOR}::qmllint)
+else()
+    # See if it's on the PATH
+    find_program(QmlLint_EXECUTABLE qmllint)
+    if(QmlLint_EXECUTABLE)
+        if(NOT QMLLINT_IS_WORKING)
+            # Try to fix common problems on Debian-based distros -- they provide /usr/bin/qmllint,
+            # which is a symlink to /usr/lib/x86_64-linux-gnu/qt4/bin/qmllint (or the Qt5 version of it).
+            # The actual executable is part of different package, so might not even be installed =>
+            # double-check whether qmllint is working by executing it
+            execute_process(
+                COMMAND ${QmlLint_EXECUTABLE} --version
+                RESULT_VARIABLE _qmllint_result
+                OUTPUT_QUIET ERROR_QUIET
             )
+            if(_qmllint_result EQUAL 0)
+                set(QMLLINT_IS_WORKING
+                    TRUE
+                    CACHE BOOL "Whether the found qmllint executable is actually usable" FORCE
+                )
+            endif()
         endif()
-    endif()
-    if(QMLLINT_IS_WORKING)
-        set(QmlLint_FOUND TRUE)
+        if(QMLLINT_IS_WORKING)
+            set(QmlLint_FOUND TRUE)
+        endif()
     endif()
 endif()
 
 # validate a list of qml files
 function(qml_lint)
-    if(NOT QMLLINT_EXECUTABLE OR NOT QmlLint_FOUND)
+    if(NOT QmlLint_EXECUTABLE OR NOT QmlLint_FOUND)
         return()
+    endif()
+
+    if(NOT TARGET qmllint_all)
+        add_custom_target(qmllint_all ALL COMMENT "Scan all .qml files with qmllint")
     endif()
 
     foreach(_file ${ARGN})
         get_filename_component(_file_abs ${_file} ABSOLUTE)
         add_custom_command(
-            OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${_file}.qmllint
-            COMMAND ${QMLLINT_EXECUTABLE} ${_file_abs}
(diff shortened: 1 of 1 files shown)


ratijas (approved):
+1

works for me, thanks!

dantti:
I wonder if we should still try the old way, like, if no Qt6 qmllint is found it might still find the Qt5 one, maybe it's best to only rely on the TARGET one?

Waqar144:
> maybe it's best to only rely on the TARGET one?

+1

I would assume the newer Qt versions set the properly.

winterz:
@ferdnyc I have no idea why this PR thinks you haven't agreed to the CLA.  If I look in cla-assistant I see you listed in there.

ferdnyc:
@dantti @Waqar144 

Honestly, my preference would be to dump the path logic and always use the target as well, yes. Say the word and I'll swap out this Find module in favor of a macro that just runs `Qt::qmllint`.

@winterz 

Yeah, it's weird. Not the first time. Let me try squashing pre-commit's contributions out of it.

ferdnyc:
Oh! I couldn't figure out what that `cmake-lint` error was about, because I thought it was cut off. But no, that's the entire message: "Missing COMMENT in statement which allows it". 

...That seems dumb. _Forcing_ a comment on every target, even simple order/aggregation ones, just bloats the output. But, OK, whatever...

ferdnyc:
@winterz And now it's happy. Though I can't say for certain it had anything to do with pre-commit, merely force-pushing the branch unchanged and getting it to re-scan might've worked just as well.

ferdnyc:
@dantti @Waqar144 @winterz 

Ahh, crap. We all assumed wrong. Even Qt 5.15.2 _doesn't_ have a `::qmllint` target. So we're back to still needing the path search, with Qt5. :disappointed:

ferdnyc:
(We can probably still dump the weird Debian logic, tho, considering the comment above it mentions qt4(!).)

dantti:
well CI with 5.15 was happy so I thought the TARGET option was working

ferdnyc:
> well CI with 5.15 was happy so I thought the TARGET option was working

Nah, I'd left the path discovery as a fallback, so it must've used it. That will always work correctly in CI, anyway, because the Right Version™ of `qmllint` is the only one installed. The issue @ratijas hit was the build discovering the wrong `qmllint` on a system with Qt5 and Qt6 installed side-by-side. (Like most any desktop Linux system.)

ferdnyc:
Well, since I took the time to clean up the logic for creating `qmllint` targets, when I thought I could just move it into `GammaRayMacrosInternal.cmake`, I'll migrate those enhancement back into the Find module and add them here.

Any objections if I also try jettisoning this weirdness?

```cmake
    if(QMLLINT_EXECUTABLE)
        if(NOT QMLLINT_IS_WORKING)
            # Try to fix common problems on Debian-based distros -- they provide /usr/bin/qmllint,
            # which is a symlink to /usr/lib/x86_64-linux-gnu/qt4/bin/qmllint (or the Qt5 version of it).
            # The actual executable is part of different package, so might not even be installed =>
            # double-check whether qmllint is working by executing it
            execute_process(
                COMMAND ${QMLLINT_EXECUTABLE} --version
                RESULT_VARIABLE _qmllint_result
                OUTPUT_QUIET ERROR_QUIET
            )
            if(_qmllint_result EQUAL 0)
                set(QMLLINT_IS_WORKING
                    TRUE
                    CACHE BOOL "Whether the found qmllint executable is actually usable" FORCE
                )
```

If `qmllint` isn't installed the `find_program()` should come up with nothing, and if it's the wrong version it's still gonna be able to successfully run `qmllint --version`. That check tells us nothing.

dantti:
From the comment on this check find_program does find qmllint, but it finds the link, which doesn't work beacuse it points nowhere, are you sure find_program() for a broken link still comes up with nothing?

I see no "FOLLOW_SYMLINKS" or similar at the docs so I guess it's still a valid check

ferdnyc:
> From the comment on this check find_program does find qmllint, but it finds the link, which doesn't work beacuse it points nowhere, are you sure find_program() for a broken link still comes up with nothing?

\*sigh\* I'd LIKE to say that I'm sure even Debian have updated their packaging since Qt4, and no longer install broken symlinks to uninstalled binaries -- but if they're still using qtchooser, I can't say that. So, back it goes.

ferdnyc:
Huh. Maybe it really IS pre-commit. I just force-pushed a nothing change that didn't remove pre-commit's changes from the branch. (I made an edit, made a second to revert it, and squashed it with my earlier commit.) ..And it still claims I haven't signed the CLA.

But earlier, when I squashed pre-commit's changes into my own and force-pushed, that immediately turned the CLA check green.

ferdnyc:
Well, I'm gonna experimentally squash out pre-commit's changes again, let's see what happens.

ferdnyc:
Instant green light on the CLA front. Interesting.

Something seems to be _up_ between the CLAssistant and pre-commit. They have _beef_.

winterz:
you could run `pre-commit install` in your gammaray checkout.  then you'll be prompted to fix (or the problems are sometimes auto-fixed for you) any problems beforehand.

dantti:
LGTM can we merge?

ferdnyc:
@winterz 
> you could run `pre-commit install` in your gammaray checkout. then you'll be prompted to fix (or the problems are sometimes auto-fixed for you) any problems beforehand.

More of a workaround than a fix, but good idea while CLAssistant and pre-commit are warring.

Tho i'm sort of tempted to _not_ — or at least, not quite yet — so I can black-box test a few more aspects of the problem.

(Like: If I push _my own_ commits to a branch **after** pre-commit, does that clear up the CLA issue? Is it an issue with the mere _presence_ of pre-commit commits anywhere on the branch? Or is it only a problem if the pre-commit commit is at the branch HEAD?)

PR #896, opened 2023-12-26
CI: Create CI presets

dantti:


ferdnyc:
Oh, that's super smart!  Makes a lot of sense.

You know, in theory all of this could move in there as well, as `testPresets`:

https://github.com/KDAB/GammaRay/blob/8a071f9a62d3e5bad82dd47e826f692ef69e3953/.github/workflows/build.yml#L111-L123

Then you could be hierarchical about the exclusions.

* A base definition (because everything excludes `quicktexturetest` and `timertoptest`)
* Inherited by a definition for each major version (or just Qt6, because all Qt6 excludes `modelinspectorttest` -- I don't see any immediate Qt5 commonalities)
* Inherited by definitions for each OS + version combo, with the remaining configs.

You could select the preset with `ctest --preset ci-${{ runner.os }}-${{ matrix.cofig.qt_major }}` (I guess the `qt_major` var would have to go back in.)

dantti:
yes, this is how I did for KDDockWidgets, but there I don't need to exclude tests, and we are trying to fix them. If you want to make a small proposal for the test presets changes we can see if it get's nicer than this explicit exclusion

CMakePresets.json
```diff
@@ -127,6 +140,102 @@
                 "CXX": "clazy",
                 "CCACHE_DISABLE": "ON"
             }
+        },
+        {
+            "name": "ci-dev-client-only-qt5",
+            "displayName": "ci-dev-client-only-qt5",
+            "description": "Qt5 client-only build",
+            "binaryDir": "${sourceDir}/build-ci-dev-client-only-qt5",
+            "generator": "Ninja",
```
ferdnyc, line 149: I was just reading the [cmake-presets docs](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#macro-expansion) -- they are _clever_ sometimes.

> All macros are evaluated in the context of the preset being used, even if the macro is in a field that was inherited from another preset. For example, if the `Base` preset sets variable `PRESET_NAME` to `${presetName}`, and the Derived preset inherits from `Base`, `PRESET_NAME` will be set to `Derived`.

AFAICT that means you could just set `"binaryDir"` to `"${sourceDir}/build-${presetName}"` once in the `base` config, and you're done.

(Tho that might actually make things harder for any `testPresets` configs I might propose, since it would mean there would have to be a 1:1 correlation of `buildPresets` names and `testPresets` names ...)

ferdnyc: > (Tho that might actually make things harder for any `testPresets` configs I might propose, since it would mean there would have to be a 1:1 correlation of `buildPresets` names and `testPresets` names ...)

Actually, I might be wrong. They say macros expand in the current context, but also that `testPresets` and `buildPresets` both inherit their build dir from the `configurationPreset` --- so that should mean the `binaryDir` name expansion happens with the name of the `configurationPreset` as `presetName`? Hmm. I'll have to experiment a bit.

PR #903, opened 2024-01-02
Remove support for Qt6 versions less than 6.3

winterz:


dantti:
are we dropping support for Qt5.15 too?

Waqar144:
> are we dropping support for Qt5.15 too?

no

Waqar144 (approved):

PR #905, opened 2024-01-03
Add KDSME as submodule for CI testing

dantti:


3rdparty/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# This file is part of GammaRay, the Qt application inspection and manipulation tool.
+#
+# SPDX-FileCopyrightText: 2010 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
winterz, line 3: change the copyright year to 2024

winterz (changes requested):
a few nitpicks. otherwise looks good

winterz (changes requested):
please change GAMMARAY_WITH_KSME to GAMMARAY_WITH_KDSME everywhere

plugins/statemachineviewer/CMakeLists.txt
```diff
@@ -46,22 +46,8 @@ if(NOT GAMMARAY_CLIENT_ONLY_BUILD)
...
-    URL "https://github.com/KDAB/KDStateMachineEditor"
-    DESCRIPTION "KDAB State Machine Editor framework"
-    TYPE RECOMMENDED
-    PURPOSE "Graphical state machine debugging."
-)
-find_package(${KDSME_PACKAGE_NAME} 1.2 CONFIG QUIET)
-
 # UI part
-if(GAMMARAY_BUILD_UI AND ${KDSME_PACKAGE_NAME}_FOUND)
+if(GAMMARAY_BUILD_UI AND GAMMARAY_WITH_KDSME)
```
ferdnyc, line 50: This breaks the ability to use a KDSME package installed separately on the system, doesn't it? Since `find_package()` is no longer even attempted, even if KDSME is installed, it won't be used — the submodule is the only option. And if the `.git` directory isn't present (e.g. package releases), there's no way to make it use a KDSME at all.

Distro packagers won't like that, for one. ...Unless I'm completely missing something?

CMakeLists.txt
```diff
@@ -250,6 +254,7 @@ endif()
 gammaray_option(GAMMARAY_INSTALL_QT_LAYOUT "Install into Qt directory layout." OFF)
 gammaray_option(GAMMARAY_MULTI_BUILD "Build multiple applicable probe configurations." ON)
 gammaray_option(GAMMARAY_BUILD_CLI_INJECTOR "Build command line injector on Windows." ON)
+gammaray_option(GAMMARAY_WITH_KDSME "Enable State Machine UI with KDSE" OFF)
```
ferdnyc, line 257:
```suggestion
gammaray_option(GAMMARAY_WITH_KDSME "Enable State Machine UI with KDSME" OFF)
```

plugins/statemachineviewer/CMakeLists.txt
ferdnyc: (FTR, my suggestion would be to keep the `find_package()`, but insert a condition to trigger the submodule stuff:

```cmake
if(NOT ${KDSME_PACKAGE_NAME}_FOUND AND GAMMARAY_WITH_KDSME)
```

Maybe even change the option name to `GAMMARAY_USE_INTERNAL_KDSME` or something.

dantti: Yes, it breaks, I did that because of the includes.
- We could include it on releases, remove/disable the git submodule check, and would also make more sense to static link to it.
- Another option would be to #ifdef the includes
- Fix KDSME includes, which isn't trivial
@winterz @Waqar144 thoughts?

ferdnyc: A fourth option:

* Keep the submodule, but instead of using it as a build `subdirectory()`, have the CI build-and-install KDSME from there before it configures GammaRay. Then pick it up with `find_package()` as usual.

The built KDSME could even be added to the GitHub cache, to speed things up.

ferdnyc: Or, in fact, you could use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) to incorporate the submodule'd KDSME — with CMake 3.24+, if you give `FetchContent_Declare()` a `FIND_PACKAGE_ARGS` argument, it'll automatically automatically try a `find_package()` first, and use the installed package if it's found.

(`FetchContent_Declare()` doesn't _have_ to download it from a remote repo. Like `ExternalProject_Add()`, you can give it a `SOURCE_DIR` argument with the path to existing sources.)

dantti: FetchContent doesn't solve the include issue, it's just a little better ExternalProject_Add() but we are avoiding it in favor of submodules.

winterz:
the distro packagers won't care because the released tarball will include the submodule. If I do things correctly when I make the release

winterz (approved):

PR #908, opened 2024-01-07
Fix two CMake uninitialized-variable warnings

ferdnyc:
- Ensure `GAMMARAY_PROBE_ABI_POSTFIX` is always set, even if it's empty.
- Use `QT_VERSION_MAJOR` in plugins/qmlsupport instead of `QT_MAJOR_VERSION`, to match rest of the CMake code.


dantti:
Since we now require Qt 5.15 we could instead just use versionless targets

winterz:
I'm merging this as it is a proper fix.
we can change to versionless if we want in a new PR

PR #909, opened 2024-01-09
Fixes for static checks

dantti:


winterz:
do the CMakeLists.txt changes belong to fixes for static checks?

otherwise looks good to me.  I think its ok to modify the 3rdparty files.

dantti:
The cmakelist changes were that Qt creator could not find the files I was looking for to fix because they were not in cmakelist.txt

Waqar144:
@dantti can you split the cmake changes in a different commit? Otherwise looks good

PR #914, opened 2024-01-12
Add header files to CMake so IDEs can list/index them

dantti:


redstrate:
Looks like the windows-specific headers is wrong somewhere?

```
CMake Error at launcher/core/CMakeLists.txt:78 (add_library):
Cannot find source file:
probeabidetector_win.h
Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm .h
.hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .hip .ispc
```

PR #916, opened 2024-01-12
Fix QT_MAJOR_VERSION to QT_VERSION_MAJOR

dantti:

PR #920, opened 2024-01-14
quickinspector: Use QT_VERSION_MAJOR

ferdnyc:
In ac8ed0c73 I fixed this in `plugins/qmlsupport/CMakeLists.txt`, but I somehow missed the exact same issue in `plugins/quickinspector/CMakeLists.txt`.

Qt target names should be constructed using `Qt${QT_VERSION_MAJOR}::target`, not `Qt${QT_MAJOR_VERSION}::target`, to conform with the rest of the build. (`QT_MAJOR_VERSION` is an ECM thing, and <s>is only found</s> should only be found in the files in `cmake/ECM/modules/`.)

### DIFF
--- a/plugins/quickinspector/CMakeLists.txt
+++ b/plugins/quickinspector/CMakeLists.txt
@@ -73,7 +73,7 @@ if(NOT GAMMARAY_CLIENT_ONLY_BUILD)
         ${gammaray_quickinspector_srcs}
     )
 
-    if(TARGET Qt${QT_MAJOR_VERSION}::Quick)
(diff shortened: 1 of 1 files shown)


dantti:
This is a duplicate of #916 but was not commited due internal conversation that ECM uses the current form, so far both options work, maybe due qt versionless targets or that this doesn't matter, still CI is happy so I would merge as other files use this form.

ferdnyc:
Never mind, @dantti already caught this in #916.

ferdnyc:
@dantti Heh. Crossed streams, I discovered your PR at the exact moment you commented.

PR #923, opened 2024-01-17
Test probe only

ferdnyc:
Way back in #765, I submitted some patches to fix the `GAMMARAY_CLIENT_ONLY_BUILD`, which had stealth-broken because it wasn't being tested in CI.

The same is now true of `GAMMARAY_PROBE_ONLY_BUILD`, so to avoid the possibility of similar stealth breakage, this PR adds test runs for probe-only builds to the standard CI, by first adding configurations for them to the `CMakePresets.json`.

Along the way, it makes some adjustments to the existing configuration(s), to facilitate adding additional CI build configurations

### Presets file `CMakePresets.json`
1. The presets file is adjusted to make better use of inheritance, removing duplicate manual settings of `"generator"` or `"binaryDir"` when those can simply be inherited.
2. To keep the file more streamlined, the new configs (and any they depend on) take advantage of a CMake feature that permits a string value for `"inherits"`, rather than a list value, when only one configuration is being inherited from.

(Not all configurations were adjusted, mostly just the ones I was touching anyway. But in both cases, they all _could_ be changed the same way.)

### CI configuration `build.yml`
1. To control whether (and which type) of unit tests are run, a new `matrix.config.tests_with` value is _optionally_ added to matrix members, set to either `qt5` or `qt6`. 
    This allows the complicated version checks:
    ```yml
    startsWith(matrix.config.qt_version, '6.')
    ```
    to be replaced with a simple:
    ```yml
    matrix.config.tests_with == 'qt6'
    ````
    and (along with `runner.os`) not only controls _which_ tests are run, but its _absence_ will **prevent** tests from being run at all. (Mostly because unit tests on a `PROBE_ONLY` build are simultaneously both pointless and redundant.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,15 +26,25 @@ jobs:
         config:
           - preset: ci-dev-client-only-qt5
             qt_version: "5.15"
+            tests_with: qt5
 
           - preset: ci-dev-client-only-qt6
             qt_version: "6.5.*"
             qt_modules: qtshadertools qtscxml
+            tests_with: qt6
 
           - preset: ci-dev-client-and-ui-qt5
             qt_version: "5.15"
+            tests_with: qt5
 
           - preset: ci-dev-client-and-ui-qt6
+            qt_version: "6.5.*"
+            tests_with: qt6
+
+          - preset: ci-dev-probe-only-qt5
+            qt_version: "5.15"
+
+          - preset: ci-dev-probe-only-qt6
             qt_version: "6.5.*"
 
     steps:
@@ -74,13 +84,13 @@ jobs:
         run: cmake --build ./build-${{ matrix.config.preset }}
 
       - name: Enable gdb attaching
-        if: ${{ runner.os == 'Linux' }}
+        if: ${{ runner.os == 'Linux' && matrix.config.tests_with != '' }}
         run: echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
 
         # Exclude
         # quickmaterialtest|quicktexturetest fails because of QT_QUICK_BACKEND=software
       - name: Run tests on Linux Qt5 (offscreen)
-        if: ${{ runner.os == 'Linux' && startsWith(matrix.config.qt_version, '5.') }}
+        if: ${{ runner.os == 'Linux' && matrix.config.tests_with == 'qt5' }}
         run: >
           ctest --test-dir ./build-${{ matrix.config.preset }} --output-on-failure
           --exclude-regex "quickmaterialtest|quicktexturetest"
@@ -92,7 +102,7 @@ jobs:
         # quickmaterialtest|quicktexturetest fails because of QT_QUICK_BACKEND=software AND QT_QPA_PLATFORM=offscreen
         # quickinspectortest|quickinspectortest2 fails at CI, local with 6.2.4 passes
       - name: Run tests on Linux Qt6 (offscreen)
-        if: ${{ runner.os == 'Linux' && startsWith(matrix.config.qt_version, '6.') }}
+        if: ${{ runner.os == 'Linux' && matrix.config.tests_with == 'qt6' }}
         run: >
           ctest --test-dir ./build-${{ matrix.config.preset }} --output-on-failure
           --exclude-regex "quickmaterialtest|quicktexturetest|quickinspectortest|quickinspectortest2"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -53,7 +61,7 @@ jobs:
         uses: turtlesec-no/get-ninja@main
 
       - name: Make sure MSVC is found when Ninja generator is in use
-        if: ${{ runner.os == 'Windows' }}
+        if: runner.os == 'Windows'
```
dantti, line 64: I know this looks nicer, but someone might be tempted to write: `if : runner.os != 'Windows' ` which is not valid yaml syntax, so for consistency it's better to keep the escaping tags

.github/workflows/build.yml
```diff
@@ -26,17 +26,25 @@ jobs:
         config:
           - preset: ci-dev-client-only-qt5
             qt_version: "5.15"
+            tests: qt5
```
dantti, line 29: While I don't like much of the idea of having a "duplicated" value to say what `qt_version` has, I would only argue that it would be more readable as `qt_version_major: 6` and also would add this for all configs.

dantti (changes requested):

.github/workflows/build.yml
ferdnyc: Well, the thing about adding it for all configs is, the intent is that unit tests will be skipped if `matrix.config.tests` isn't set at all. (Hence, also, the name.)

.github/workflows/build.yml
ferdnyc: Aha! I missed that. I'll fix those up.

.github/workflows/build.yml
dantti: right, I forgot you mentioned that, maybe `tests_with` is more meaningful?

ferdnyc: So it's basically a tristate: `qt5` == "Run the Qt5 tests", `qt6` == "Run the Qt6 tests", unset == "run no tests".

ferdnyc: I currently have it set only on the client-only builds, because I don't think any of the unit tests require the UI, do they? (If they do, I can add/move it to the client-and-ui builds.)

ferdnyc: Sorry, hadn't seen your reply, GitHub is being weird today. Sure, I'll change it to `tests_with`.

dantti: iirc there are many unit tests that require ui

ferdnyc: Hm. Weird, they all pass in a client-only build!

ferdnyc: Possible the CMake config is auto-excluding the UI ones, though.

.github/workflows/build.yml
```diff
@@ -65,19 +73,19 @@ jobs:
...
       - name: Install dependencies on Ubuntu
-        if: ${{ runner.os == 'Linux' }}
+        if: ${{ runner.os == 'Linux' && matrix.config.tests_with != "" }}
         run: |
           sudo apt update -qq
           sudo apt install -y gdb
 
       - name: Enable gdb attaching
-        if: ${{ runner.os == 'Linux' }}
+        if: ${{ runner.os == 'Linux' && matrix.config.tests_with != "" }}
```
dantti, line 82: Must use single quotes

ferdnyc: Argh! Gets me every time, thanks.

.github/workflows/build.yml
ferdnyc: ^ Yup, it sure was. 49 tests with UI enabled, 14 without.

ferdnyc: So, I've now enabled tests for both client-only and client-and-ui builds. Though running them for client-only may be unnecessary, since they're just a subset of the tests that will all be run for client-and-ui.

dantti (approved):
LGTM

winterz:
@ferdnyc would you fix the conflicts please.
I'm sorry we forgot to merge this one and now we have a few conflicts

ferdnyc:
@winterz Sorry for the delay, will do.

ferdnyc:
@winterz Done. I took the opportunity to squash some of the back-and-forth away, since I had to rewrite all of the commits to rebase the branch.

winterz:
the 2 CI failures are related to KDStateMachineEditor dependency which we're still working.

PR #925, opened 2024-01-17
CI: Use latest stable Qt6.6 and next beta 6.7

dantti:


dantti:
We likely need a new aqt version to be able to install 6.7.0 on linux since Qt arch for linux has been renamed to linux_gcc64

PR #930, opened 2024-01-17
CMake: Split Library and Plugin types into PROBE and UI; support BUILD_SHARED_LIBS

Kentzo:
User's choice of what's to be built is disambiguated from how it's built. Both static and shared builds of both library and plugins is now supported.


Kentzo:
WIP and open for discussion.

Kentzo:
The idea was to allow static probes alongside non-static UI. But that seems to be too much work: pretty much every library needs to be split into static and shared version with actual build path determined by config variables. That's doable, but probably too much effort. What's needed is two variants of the project:

1. UI and shared probes for runtime injection
2. No UI but a static probe library for compile-time injection

For [2] some modifications are still necessary to make sure that static probe libraries carries all dependencies in itself.

PR #934, opened 2024-01-18
CMake: Explicitly handle incompatible option overrides

Kentzo:
Provide sensible defaults for the *ONLY* builds and explicitly verify incompatible configurations.


Kentzo:
The goal here is to have explicit errors when user sets options to conflicting values.

Kentzo:
Test failure seems to be unrelated to the changes.

PR #935, opened 2024-01-19
Feature: Ctrl+LeftClick Drag panning for non-pan tools

### LINKED ISSUE #933: Quick Scenes: always pan view with <modifier>+click
Currently one has to switch to the pan view tool then click and drag to move the preview of the app.
It would be nice, if regardless of the tool selected, one could just hold a modifier key + left/right click to move the preview.

mattkdab:
Requested feature.

Enable Ctrl+Shift+LeftClick Dragging to pan the remote view when using interaction modes other than ViewInteraction.

Should close #933

### DIFF
--- a/ui/remoteviewwidget.cpp
+++ b/ui/remoteviewwidget.cpp
@@ -1013,8 +1013,7 @@ void RemoteViewWidget::mousePressEvent(QMouseEvent *event)
     if (m_interactionMode != NoInteraction
         && m_interactionMode != ViewInteraction
         && event->buttons() & Qt::LeftButton
-        && event->modifiers() & Qt::ControlModifier
-        && event->modifiers() & Qt::ShiftModifier) {
+        && event->modifiers() & Qt::ControlModifier) {
 
         pan();
     } else {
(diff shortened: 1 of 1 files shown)


mattkdab:
just noticed #741 is sort of the same as this?

g-fb:
Any reason for requiring both Ctrl and Shift to be pressed?

Anyway, thank you for adding this.

mattkdab:
> Any reason for requiring both Ctrl and Shift to be pressed?
> 
> Anyway, thank you for adding this.

Hm, right I suppose it _could_ just be Ctrl, since this is only when not using the pan tool (Ctrl clicking with pan tool performs a Pick Element). I'll update with that change.

Waqar144:
related https://github.com/KDAB/GammaRay/pull/741

Waqar144 (approved):

winterz:
@mattkdab I'm stumped why the cla-assistant doesn't recognize you.  I'll approve this and merge.

winterz (approved):

PR #936, opened 2024-01-19
Build static probe without internal dependencies

Kentzo:
Adjust types of the probe dependency tree so that the final static library has no internal dependencies. Install headers, pri and cmake files.


Kentzo:
I will split this PR into smaller pieces.

PR #938, opened 2024-01-24
CMake: Add the INSTALL_DIR argument to gammaray_add_plugin

Kentzo:
Use this new feature to define the gammaray_geopositioninfosource plugin via the gammaray_add_plugin rather than custom add_library.


Kentzo:
Failed tests will get resolved after https://github.com/KDAB/GammaRay/pull/940 is merged.

dfaure-kdab:
#940 was closed, so what should happen to this? Surely it can't go in with failing tests...

PR #939, opened 2024-01-24
CMake: Build gammaray_common_internal and gammaray_lz4 as OBJECTs

Kentzo:
This allows avoid installation and export of these intermediate libraries when building the STATIC probe.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -23,7 +23,7 @@ configure_file(
 )
 
 add_library(
-    gammaray_lz4 STATIC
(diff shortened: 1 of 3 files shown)


ferdnyc (commented):
This makes a lot of sense to me. `OBJECT` libraries are an underused feature of CMake, and as long as they're built with PIC enabled (which these already were, even as `STATIC` library components), it's completely sensible to construct either `STATIC` or `SHARED` output libraries which include other `OBJECT`-library components, reducing the installable-file count.

redstrate (approved):

PR #940, opened 2024-01-24
CMake: Define gammaray_pch_core_gui as an INTERFACE library

Kentzo:
INTERFACE avoids actual compilation and more complex dependency relation of a STATIC library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,9 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_LINK_DEPENDS_NO_SHARED ON)
 
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+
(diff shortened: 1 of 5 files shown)


Waqar144:
This will make every target have its own PCH resulting in a build that is slower than non-pch build plus a massive increase in the build dir size.

Kentzo:
Thank you for taking a look.

I was inspired by the following excerpt from the [target_precompile_headers](https://cmake.org/cmake/help/latest/command/target_precompile_headers.html) doc:

> A notable exception to this is where an [interface library](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#interface-libraries) is created to define a commonly used set of precompile headers in one place and then other targets link to that interface library privately. In this case, the interface library exists specifically to propagate the precompile headers to its consumers and the consumer is effectively still in control, since it decides whether to link to the interface library or not.

Comparison of running times of GitHub Actions suggests that the compilation time indeed increased. E.g. https://github.com/KDAB/GammaRay/actions/runs/7637161294/job/20805489439?pr=941 vs https://github.com/KDAB/GammaRay/actions/runs/7645378040/job/20831803946?pr=940.

I will investigate further.

Kentzo:
You were correct, I misunderstood the doc.

PR #947, opened 2024-02-08
Remove support for Qt IVI

mattkdab:


winterz:
do we need to also remove tests/qtivipropertymodeltest.cpp?
I also see docs/manual/gammaray-manual.qdoc and docs/manual/gammaray-tools.qdoc refert to QtIVI

please add a blurb in the CHANGES file under the "Version 3.1.0" section that support for QtIVIfoo is removed

winterz (approved):

PR #951, opened 2024-02-26
Resolve "unregistered datatype 'QHeaderView*'"

### LINKED ISSUE #949: QMetaMethod::invoke:  'QHeaderView*'
I get the following Error on the Console: 

`QMetaMethod::invoke: Unable to handle unregistered datatype 'QHeaderView*'`

Doesn't seem to be very urgent as I can continue to use GammaRay. 
But I wanted to report the issue, however.

mattkdab:
Fix #949
When running GR from command line we get:
```
QMetaMethod::invoke: Unable to handle unregistered
datatype 'QHeaderView*'
```
on tab loading. Register the type to fix.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -24,6 +24,8 @@
 #include <QApplication>
 #include <QStringList>
 
+#include <QHeaderView>
+
 using namespace GammaRay;
 
 int main(int argc, char **argv)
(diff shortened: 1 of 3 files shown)


mattkdab:
I'm uncertain that this is the best location to put this code.

Waqar144:
Can be put in UIStateManager since that's where the error is coming from

PR #957, opened 2024-03-04
Fix build on 6.7 for after QDeferredDeleteEvent export reversion

mattkdab:
In the next 6.7 release, this reversion will be done:
https://codereview.qt-project.org/c/qt/qtbase/+/544223
which will break building GammaRay against 6.7
Use static_cast in an explicit specialization
GammaRay::DynamicCast<QDeferredDeleteEvent *, QEvent>
to avoid missing typeinfo errors from linker


mattkdab:
@redstrate

redstrate (changes requested):
please change the commit messages to be a bit more descriptive for later readers. Something like this, feel free to change:

[Fix build on 6.7 for after QDeferredDeleteEvent export reversion](https://github.com/KDAB/GammaRay/pull/957/commits/0a6c0c9c77d4178faa018081ed8b28b3105ee345)

* In Qt 6.7 QDeferredDeleteEvent is now inaccessible, but we can work around it by using a `static_cast`. We also need to start linking to Qt::CorePrivate for `<private/qcoreevent_p.h>` in 6.7.

Will probably want to squash too, once everything is done

mattkdab:
@redstrate you should be able to squash-merge once this is approved

redstrate (approved):
LGTM now, @Waqar144 could you do a sanity check? to make sure the formatting and all looks fine to you

plugins/quickinspector/quickinspector.cpp
```diff
@@ -110,7 +110,6 @@ Q_DECLARE_METATYPE(QSGNode::Flags)
 Q_DECLARE_METATYPE(QSGNode::DirtyState)
 Q_DECLARE_METATYPE(QSGGeometry *)
 Q_DECLARE_METATYPE(QMatrix4x4 *)
-Q_DECLARE_METATYPE(const QMatrix4x4 *)
```
Waqar144, line 113: unrelated?

core/typetraits.h
```diff
@@ -27,6 +32,14 @@ Out DynamicCast(In *in)
     return dynamic_cast<Out>(in);
 }
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+template<>
+inline QDeferredDeleteEvent *DynamicCast<QDeferredDeleteEvent *, QEvent>(QEvent *in)
+{
+    return static_cast<QDeferredDeleteEvent *>(in);
```
Waqar144, line 39: This will not work if UBSAN is being used I think.

core/metaproperty.h
```diff
@@ -15,6 +15,7 @@
 #define GAMMARAY_METAPROPERTY_H
 
 #include "gammaray_core_export.h"
+#include <common/metatypedeclarations.h>
```
Waqar144, line 18: related or?

plugins/quickinspector/quickinspector.cpp
mattkdab: this was previously causing a multiple declaration but doesnt appear to be doing so now, i can revert this.

core/metaproperty.h
mattkdab: also could not build without this but no longer seems to be an issue in the current commit. can revert

mattkdab: reverted

plugins/quickinspector/quickinspector.cpp
mattkdab: reverted

Waqar144 (approved):

redstrate (approved):

Issue #954, opened 2024-02-28
Can't attach to PySide2 Application importing QtCore


brycegbrazen:
Hi all,

I'm writing a Python application using PySide2, and when I add an import of:

```from PySide2 import QtCore```

to the top of the file, Gamma Ray no longer lists the application on the "Attach" tab (and therefore cannot attach to the process). 

Does anyone know why this is the case? Is this a bug?

Here's a tiny example file that shows the issue:
[simple_test.zip](https://github.com/KDAB/GammaRay/files/14439346/simple_test.zip)

# Specs

- Windows 10
- Qt version 5.15.2
- PySide2 PyPi package version 5.15.2.1
- Gamma Ray 3.0.0

**Comment by Waqar144** [comment]
Please try with latest master

**Comment by brycegbrazen** [comment]
@Waqar144 I rebuilt Gamma Ray with latest master [commit](https://github.com/KDAB/GammaRay/commit/abd80cc20acfd85c0509121736931fd5a62b6a25)  and the application still does not appear in the list of processes that Gamma Ray can attach to after adding the `from PySide2 import QtCore` line.

**Comment by Waqar144** [comment]
@brycegbrazen Can you share what are the contents of your PySide2 package. For e.g., on linux on my pc `/usr/lib/python3.11/site-packages/PySide6/` has 

```
/usr/lib/python3.11/site-packages/PySide6/QtConcurrent.abi3.so
/usr/lib/python3.11/site-packages/PySide6/QtCore.abi3.so
... more .so files
```

Also, if possible put a std::cout in `ProbeABIDetector::containsQtCore(const QByteArray &line)` and print out the line when trying to attach to a pyside app. It will make it easier to figure out what's going wrong.

**Comment by brycegbrazen** [comment]
@Waqar144 Inside of the PySide2 package I've got some similar files to the ones you mentioned above but with the `.dll` and `.lib` extensions:
```
pyside2.abi3.dll
pyside2.abi3.lib
... more .dll files
```
I do also see a .dll file specifically for Qt5Core: `Qt5Core.dll`

brycegbrazen:
Unfortunately I'm not sure how to build and launch Gamma Ray in order to see any `std::cout` print statements I add, so I cannot do that part without some guidance.

Waqar144:
@brycegbrazen 

[...]

PR #961, opened 2024-03-20
Fix probe not picking up PySide2 applications on some corner cases

aambrosano:
Fixes issue #954

As per comment in patch, sometimes GammaRay doesn't pick up PySide2 applications on Windows.
This seems to happen when QtCore is imported before other Qt modules.

The reason for that is that in that case QtCore.pyd will appear before the QtCore DLL in the list of loaded modules, and so it will be picked up instead of the DLL for the version check.

As the pyd file doesn't have any information about the Qt version among its file details, the version lookup will fail.

Also, since all pyd files will in turn load a corresponding DLL we can safely skip all pyd modules when navigating through loaded modules.

### DIFF
--- a/launcher/core/probeabidetector.cpp
+++ b/launcher/core/probeabidetector.cpp
@@ -151,6 +151,14 @@ static bool checkQtCoreSuffix(const QByteArray &line, int index)
         return false;
     }
 
+    // Sometimes when using PySide2 on Windows we end up loading QtCore.pyd, which doesn't
+    // have information about the corresponding Qt version among its file details.
+    // We can safely skip it since that will in turn load a QtCore dll file with the right
+    // information, and that will be picked up by this function.
+    if (line.endsWith(".pyd")) {
+        return false;
+    }
+
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


launcher/core/probeabidetector.cpp
```diff
@@ -161,6 +161,13 @@ bool ProbeABIDetector::containsQtCore(const QByteArray &line)
     // Windows Qt[X]Core[d].dll
 
     for (int index = 0; (index = line.indexOf("Qt", index)) >= 0; ++index) {
+        // Sometimes when using PySide2 on Windows we end up loading QtCore.pyd, which doesn't
+        // have information about the corresponding Qt version among its file details.
+        // We can safely skip it since that will in turn load a QtCore dll file with the right
+        // information, and that will be picked up by this function.
+        if (line.endsWith("pyd"))
```
Waqar144, line 168: Can this check be done in checkQtCoreSuffix?

Waqar144:
A test case for this in https://github.com/KDAB/GammaRay/blob/master/tests/probeabidetectortest.cpp#L85 would be good

launcher/core/probeabidetector.cpp
aambrosano: Actually changed to skip pyd files altogether from `qtCoreForProcess`

aambrosano: Moved back to `checkQtCoreSuffix` (although it seems a bit pessimizing), this way we can also easily test that it works through `probeabidetectortest`.

PR #962, opened 2024-03-21
Fixing Bug where a childs in child in a zero size element are not found.

53845714nF:
I have the Problem by using the Picker, they didn't find elements in elements in  a element which are zero size. 

For example:
```
import QtQuick 2.15
import QtQuick.Window 2.15

Window {
    width: 640
    height: 480
    visible: true
    title: qsTr("Minimal Pick Fail")

    Rectangle {
        id: zerosizereact1
        width: 0
        height: 0

        Rectangle {
            id: zerosizereact2
            width:0
            height: 0

            Rectangle {
                id: greenrectangle
                width: 100
                height: 100
                color: "green"
            }
        }
    }
}
```
When I click the green rectangle, the Picker chose the hole screen as Element.
grafik

### DIFF
--- a/plugins/quickinspector/quickinspector.cpp
+++ b/plugins/quickinspector/quickinspector.cpp
@@ -799,6 +799,30 @@ void QuickInspector::pickElementId(const GammaRay::ObjectId &id)
         m_probe->selectObject(item);
 }
 
+QRectF QuickInspector::combinedChildrenRect(QQuickItem *parent) const
+{
+    auto rect = parent->childrenRect();
+
+    const auto childItems = parent->childItems();
+    for (const auto child : childItems) {
+        auto childRect = child->childrenRect();
+
+        // Get Global positon of childRect
+        QPointF childGlobalPos = child->mapToScene(QPointF(0, 0));
+
+        // Convert global position to local coordinates of the parent object
+        QPointF localChildPos = parent->mapFromScene(childGlobalPos);
+
+        // Adjust childRect to be in local coordinates of the parent object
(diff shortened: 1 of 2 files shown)


redstrate:
> Does the failed test have something to do with me? I don't see what's the problem.

It's just a faulty test, sorry. I did re-run the failed pipeline so now it's all green

plugins/quickinspector/quickinspector.cpp
```diff
@@ -799,6 +799,29 @@ void QuickInspector::pickElementId(const GammaRay::ObjectId &id)
         m_probe->selectObject(item);
 }
 
+QRectF QuickInspector::combinedChildrenRect(QQuickItem *object) const
+{
+    auto rect = object->childrenRect();
+
+    for (const auto child : object->childItems()) {
```
Waqar144, line 806: put `object->childItems()` into a separate variable to avoid a detach warning
```c++
   const auto childItems = object->childItems();
   for (const auto child : childItems) {
```

plugins/quickinspector/quickinspector.cpp
```diff
@@ -799,6 +799,29 @@ void QuickInspector::pickElementId(const GammaRay::ObjectId &id)
         m_probe->selectObject(item);
 }
 
+QRectF QuickInspector::combinedChildrenRect(QQuickItem *object) const
```
Waqar144, line 802: can you rename `object` to `parent`? Increases the readibility a lot imo

53845714nF:
I have added the changes.

Waqar144:
Okay, thanks.

PR #965, opened 2024-03-26
CMake: Use qtpaths -query over qmake -query when possible

### LINKED ISSUE #964: Building against custom Qt6 build fails if `qmake` is missing
Hi there, we do our own Qt builds[^1] for [OBS Studio](https://github.com/obsproject/obs-studio), which do not include qmake. It seems that it's only used for the `qmake -query` functionality which can also be provided by `qtpaths -query`.

I ended up applying the following patch to make the build work with our Qt builds:
```diff
diff --git a/cmake/KDAB/modules/KDQtInstallPaths.cmake b/cmake/KDAB/modules/KDQtInstallPaths.cmake
index 4fa7d04b9..ec1729497 100644
--- a/cmake/KDAB/modules/KDQtInstallPaths.cmake
+++ b/cmake/KDAB/modules/KDQtInstallPaths.cmake
@@ -16,8 +16,8 @@ if(NOT DEFINED QT_VERSION_MAJOR)
     message(FATAL_ERROR "Please set QT_VERSION_MAJOR first (ie. set(QT_VERSION_MAJOR 5))")
 endif()

-if(TARGET Qt${QT_VERSION_MAJOR}::qmake)
-    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::qmake LOCATION)
+if(TARGET Qt${QT_VERSION_MAJOR}::qtpaths)
+    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::qtpaths LOCATION)
 else()
     message(FATAL_ERROR "No supported Qt version found. Make sure you find Qt before calling this")
 endif()
```

I'm not 100% sure if this could be PR'd as-is (with some variable name changes) as some comments lead me to believe that this would not be compatible with Qt5? I'm also not sure if this would break any of the optional features I do not have enabled in my current build.

[^1]: https://github.com/obsproject/obs-deps

mattkdab:
Fix #964

### DIFF
--- a/cmake/KDAB/modules/KDQtInstallPaths.cmake
+++ b/cmake/KDAB/modules/KDQtInstallPaths.cmake
@@ -16,16 +16,11 @@ if(NOT DEFINED QT_VERSION_MAJOR)
     message(FATAL_ERROR "Please set QT_VERSION_MAJOR first (ie. set(QT_VERSION_MAJOR 5))")
 endif()
 
-# can use qtpaths to find version instead of qmake
-# but cmake target Qt5::qtpaths DNE, so use qtpaths for 6 only
-if (QT_VERSION_MAJOR EQUAL 6)
-    set(findqt qtpaths)
-else()
-    set(findqt qmake)
-endif()
-
-if(TARGET Qt${QT_VERSION_MAJOR}::${findqt})
-    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::${findqt} LOCATION)
+# use qtpaths if qmake not found
+if(TARGET Qt${QT_VERSION_MAJOR}::qmake)
+    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::qmake LOCATION)
+elseif(TARGET Qt${QT_VERSION_MAJOR}::qtpaths)
+    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::qtpaths LOCATION)
 else()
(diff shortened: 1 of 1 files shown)


cmake/KDAB/modules/KDQtInstallPaths.cmake
```diff
@@ -16,8 +16,16 @@ if(NOT DEFINED QT_VERSION_MAJOR)
...
-    get_target_property(QT_QMAKE_EXECUTABLE Qt${QT_VERSION_MAJOR}::qmake LOCATION)
+# can use qtpaths to find version instead of qmake
+# but cmake target Qt5::qtpaths DNE, so use qtpaths for 6 only
+if (QT_VERSION_MAJOR EQUAL 6)
+    set(findqt qtpaths)
+else()
+    set(findqt qmake)
+endif()
+
+if(TARGET Qt${QT_VERSION_MAJOR}::${findqt})
```
Waqar144, line 27: can we do `if` / `else if` instead of enforcing qtpaths with Qt6?

```
if (qmake)
  // ...
elseif (qtpaths)
  // ...
else
  message("qt not found")
```

mattkdab: yes good plan, checking existence for both and going with what is available is much safer.
out of curiosity, why prefer qmake over qtpaths when both are available?

Waqar144: Because that's how it was and we haven't received any issues about it besides the linked one, so I'd still keep it at the top and try `qtpaths` only if qmake didn't work

PR #968, opened 2024-04-04
Fix deserialization of graphicsitems due to shape being a QPainterPath

dfaure-kdab:
This fixes
`QVariant::load: unknown user type with name QPainterPath.`
warnings, and infinite Loading... in the scene inspector (when that property is the only problem...)


Waqar144 (approved):

dfaure-kdab:
Fixed Qt5 build by adding Q_DECLARE_METATYPE (and cleaning up the duplicate one)

PR #969, opened 2024-04-04
Fix inspection of graphics scenes with nested items

dfaure-kdab:
The SceneModel had a broken parent() implementation, it always returned row = 0 for toplevel items, instead of looking up the position in the topLevelItems() list.

While at it, make this model more robust against the fact that the order of QGraphicsItem::childItems() can change due to stackBefore() or changing the Z order, so sort that list (by pointer address) to make it stable.

There's still one way to break that model:
QGraphicsItem::setParentItem(). But that's for another day.

### DIFF
--- a/plugins/sceneinspector/scenemodel.cpp
+++ b/plugins/sceneinspector/scenemodel.cpp
@@ -106,6 +106,29 @@ int SceneModel::rowCount(const QModelIndex &parent) const
     return topLevelItems().size();
 }
 
+// QGraphicsItem::childItems() sorts them by stacking order, which takes into account the
+// items' insertion order, calls to QGraphicsItem::stackBefore(), and Z-values
+// That's too prone to change under our feet without telling us, breaking model invariants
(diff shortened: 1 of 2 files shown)


plugins/sceneinspector/scenemodel.cpp
```diff
@@ -106,16 +106,37 @@ int SceneModel::rowCount(const QModelIndex &parent) const
...
+// items' insertion order, calls to QGraphicsItem::stackBefore(), and Z-values
+// That's too prone to change under our feet without telling us, breaking model invariants
+// So always just sort them... by pointer address, for lack of a better idea
+namespace
+{
+    QList<QGraphicsItem *> sortedChildItems(QGraphicsItem *parent)
+    {
+        auto items = parent->childItems();
+        std::sort(items.begin(), items.end());
+        return items;
```
Waqar144, line 119: Might have perf issues, but I guess we'll find out.

PR #986, opened 2024-05-02
Disable GAMMARAY_WITH_KDSME on Qt 6 Linux

Cuperino:
This works around the fact that Qt6StateMachine is missing from some of the Qt 6 Linux workers, causing CI to fail.


dantti:
Closing this as I'm trying to fix CI issues at #987

PR #988, opened 2024-05-02
Use latest Qt6 version 6.7

dantti:


dantti:
6.7.0 has warnings which become errors
6.7.1 not properly available yet

biscuitcakes:
I'm trying to use GammaRay with my app which is running Qt 6.7 but it won't attach and just crashes. Would this fix it? If so is there a way I can help?

PR #994, opened 2024-06-23
Enable KDStateMachineEditor build on all OSes

dantti:
Need to wait for KDSME MR to land


dantti:
I had to change the main CMakeLists.txt please review if it makes sense

PR #995, opened 2024-06-27
Make link clickable in Yama security error window

Ashark:
It is inconvenient to copy the link manually every time and pasting it in browser. Better make it clickable in window.

Before:

before

After:

after

---

Please check if I missed some other place to replace `\n` with `<br>`.

I did not used the `errorBox.setTextFormat(Qt::MarkdownText);`, because it was removed with [this](https://github.com/KDAB/GammaRay/commit/bd6863392db92dac6130f46d5fb1e6c6ac32c1a7) commit to support building with Qt < 5.14. I do not know if this version is still needed by anyone, so I decided to use `Qt::RichText`.

### DIFF
--- a/launcher/cli/main.cpp
+++ b/launcher/cli/main.cpp
@@ -344,9 +344,9 @@ int main(int argc, char **argv)
         QMessageBox errorBox;
         errorBox.setWindowTitle("Launcher Error");
         errorBox.setIcon(QMessageBox::Icon::Critical);
-        // errorBox.setTextFormat(Qt::MarkdownText);
+        errorBox.setTextFormat(Qt::MarkdownText);
         errorBox.setTextInteractionFlags(Qt::TextBrowserInteraction);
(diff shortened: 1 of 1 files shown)


Ashark:
In INSTALL.md I see that the 5.15 is a minimum supported version:
> To build GammaRay you will need *at least*:
> - Qt 5.15 or higher, or Qt 6.3 or higher
  (GammaRay v3.0.0 was the last release to support Qt versions less than 5.15)
 
 So perhaps the current MR can just reenable the `Qt::MarkdownText`.

redstrate:
> In INSTALL.md I see that the 5.15 is a minimum supported version:
> 
> > To build GammaRay you will need _at least_:
> > 
> > * Qt 5.15 or higher, or Qt 6.3 or higher
> >   (GammaRay v3.0.0 was the last release to support Qt versions less than 5.15)
> 
> So perhaps the current MR can just reenable the `Qt::MarkdownText`.

Correct, we can use MarkdownText here since the minimum Qt was bumped sometime ago

Ashark:
> Correct, we can use MarkdownText here since the minimum Qt was bumped sometime ago

Done.

PR #997, opened 2024-07-02
Widgets and models: enable recursive filtering and keeping children

dfaure-kdab:
This makes it a lot easier to find a widget or a model in the tree, and still see all of its children.

Maybe this should be generalized to all trees with a filter lineedit?

### DIFF
--- a/core/remote/serverproxymodel.h
+++ b/core/remote/serverproxymodel.h
@@ -22,6 +22,17 @@
 #include <QVector>
 
 namespace GammaRay {
+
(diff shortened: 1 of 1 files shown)


Waqar144:
> Maybe this should be generalized to all trees with a filter lineedit?

Yes, would be a good idea.

dfaure-kdab:
> > Maybe this should be generalized to all trees with a filter lineedit?
> 
> Yes, would be a good idea.

Done, please review.

Waqar144 (approved):

PR #1015, opened 2024-10-03
Fix build error due to Qt::Gui not found during check_cxx_symbol_exists

dfaure-kdab:
I'm not sure why that alias isn't available at that point... CMake says:
Target "cmTC_4f21f" links to: Qt::Gui but the target was not found.

It works with Qt6::Gui.


dfaure-kdab:
Thanks. Can you also update master? Not sure if you want to merge or cherry-pick...

Waqar144:
forward merge, I can do that later.

PR #1026, opened 2024-12-05
Fix build with latest Qt dev

CarlSchwan:
Broken by https://codereview.qt-project.org/c/qt/qtbase/+/603678

### DIFF
--- a/core/remoteviewserver.cpp
+++ b/core/remoteviewserver.cpp
@@ -188,7 +188,7 @@ void RemoteViewServer::sendTouchEvent(int type, int touchDeviceType, int deviceC
     const QEventPoint::States states(touchPointStates);
     QTouchEvent event(QEvent::Type(type), m_touchDevice.get(), Qt::KeyboardModifiers(modifiers), states, touchPoints);
 
-    auto *mut = QMutableTouchEvent::from(&event);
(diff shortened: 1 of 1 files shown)


redstrate:
question: is qtdev now on 6.9 or 6.10? If it's 6.9, it's probably worth changing the commit description to reflect that. If not, ignore

and one nit: please add a comment to `CHANGES`

CarlSchwan:
The patch in question will be for Qt 6.9, I will edit the description

core/remoteviewserver.cpp
```diff
@@ -188,7 +188,7 @@ void RemoteViewServer::sendTouchEvent(int type, int touchDeviceType, int deviceC
     const QEventPoint::States states(touchPointStates);
     QTouchEvent event(QEvent::Type(type), m_touchDevice.get(), Qt::KeyboardModifiers(modifiers), states, touchPoints);
 
-    auto *mut = QMutableTouchEvent::from(&event);
+    auto *mut = static_cast<QMutableTouchEvent *>(&event);
```
Waqar144, line 191: Will this not break older qt versions?

Waqar144: perhaps not, the method seems to be doing a static_cast internally as well.

PR #1029, opened 2025-01-02
Fix the OpenGLScreenGrabber for QQuickWidgetOffscreenWindow in Qt6.8+

milianw:
For me at least, the y offset has to be 0, otherwise we grab an offset part of the image. Furthermore, the x offset has to be hidpi adapted too.

I cannot test the other Qt versions, so I'll assume this code used to work fine there and only touch this for Qt 6.8+


milianw:
without this patch:

Screenshot_20250102_165823

PR #1032, opened 2025-01-03
Revive the RenderModeRequest::apply for Qt6

milianw:
This was disabled in the original port in patch b1e8e7fd5b. In my own testing, it works fine nowadays, so let's re-enable it instead of silently disabling the code in a non-obvious way.


iamsergio:
was the CI failure related ?

if not rebase and we'll check again

PR #1036, opened 2025-01-17
Fix remote decoration drawing on OpenGL

dangelog:
On Qt 6.8 enabling the "decorate target" button doesn't do anything. If we have to inject custom GL commands, we have to inform QtQuick that we're doing so, so wrap our after-rendering screengrab and overlay drawing in the appropriate calls.
Also, apparently can only be done while a pass is being recorded, so switch the connect() from afterRendering (pass has been recorded and finished, but not submitted) to afterRenderPassRecording (QQ2 has drawn itself but the pass is still active).

### DIFF
--- a/plugins/quickinspector/quickscreengrabber.cpp
+++ b/plugins/quickinspector/quickscreengrabber.cpp
@@ -577,8 +577,7 @@ OpenGLScreenGrabber::OpenGLScreenGrabber(QQuickWindow *window)
 #endif
         this,
         &OpenGLScreenGrabber::windowAfterRendering,
-        Qt::DirectConnection
-    );
+        Qt::DirectConnection);
 }
 
 OpenGLScreenGrabber::~OpenGLScreenGrabber() = default;


Cuperino:
With this, realtime changes to QSG_VISUALIZE stop showing on the scene inspector, but they continue to work on the program's window. There must be something about the order of operations, or the scope being delimited by beginExternalCommands and endExternalCommands.

It might be okay to not show QSG_VISUALIZE changes in the inspector, considering there's some latency in the visuals. Tho not showing them would impact remote workflows.

dangelog:
Uhm, so it's a bit of a catch 22:

* if we want decorations to be drawn on the target, we need to do it **before** the rendering pass is complete;
* if we want to capture the special QSG drawing modes, we need to capture **after**. However, if we do this, we also end up capturing our own decorations (d'oh)...

Any idea how to solve this?

dangelog:
This is also likely to be relevant for RHI. While maybe with GL we can ignore Qt docs and capture+draw after the render pass is complete, we 100% will not be able to do so there.

Cuperino:
I'd like for us to move towards an RHI agnostic implementation at some point, so my inclination is towards capturing and drawing before the render pass is complete. The usability of looking at some of the special QSG drawing modes from the inspector is a bit questionable due to the low frame rate. Things should be good as long as users are able to view them from the application.

dangelog:
Well, that's the other problem I'm facing with the RHI implementation :)

I'd like to trigger a capture and then draw the target decorations, but I can't find a way to do so using RHI APIs. (In particular the "triggering a capture" part. I have a patch that "works", but captures after the render pass, thus it also gets the target decorations.)

PR #1055, opened 2025-05-06
cmake: Use RelWithDebInfo instead of Release for dev preset

iamsergio:
Another option would be Debug, but seems CI depends on Release.
Would need refactoring, mayeb introduce ci-base


dantti (approved):
LGTM

PR #1058, opened 2025-05-14
ci: Simplify build.yml by moving Qt5 to its own workflow

iamsergio:
By introducing minor duplication, we:

  - Make qt_version a matrix dimension, making it easy to add more Qt 6 versions

  - Remove duplication (There were separate blocks for running Qt5 tests already)

  - Make the Qt6 workflow maintainable while the Qt5 workflow can go to sleep and close shop for changes

Otherwise it's very hard to make any clean changes to this workflow


Waqar144:
This is the last version with Qt5 support, post release Qt5 support should be dropped as its EOL. We can then drop the qt5 ci as well.

PR #1062, opened 2025-05-22
drop qt5 support

Waqar144:


iamsergio:
@Waqar144 INSTALL.md needs a line saying users should use version 3.2 for Qt 5.15 support
And remove Qt 5.15 from requirements line

Issue #959, opened 2024-03-15
GammaRay: Failed to load: ..\gammaray\plugins\gammaray\3.0\qt6_5-GNU-x86_64d\gammaray_probed.dll


mml-0104:
Successfully compile and run the launcher on Windows 10 using Qt6.5.3，but **Attach** failed.
The console has the following prompt: 
![image](https://github.com/KDAB/GammaRay/assets/80971190/a4f4c1d2-cc35-45c3-87d2-03045f62561d)

There is no error reported on the **Self Test** page: 
![image](https://github.com/KDAB/GammaRay/assets/80971190/9229ffd5-b84a-4f34-a7a7-1b9aa10ac647)

 Use default cmake optionals

`cmake -G Ninja -DCMAKE_INSTALL_PREFIX="H:/Downloads/gammaray" -DCMAKE_PREFIX_PATH="D:/Qt/6.5.3/mingw_64" ../`

The system environment variable **PATH** setting: 

>C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files\dotnet\;D:\Qt\6.5.3\mingw_64\bin;D:\Qt\Tools\CMake_64\bin;D:\Qt\Tools\Ninja;D:\Qt\Tools\mingw1120_64\bin;

kpopv:
As suggested in #983, I've tried procmon to see where it is failing for me (cf. #960). In my case, gammaray_probed.dll can be found. It fails to access some paths within a target directory (cf. paths.cpp pluginPaths / addPluginPath), but I'm unsure if those are required:
- QueryOpen	plugins\gammaray\3.0\qt5_15-GNU-x86_64d\gammaray_probed.dll	SUCCESS
- QueryOpen	plugins\gammaray\3.0\qt5_15-GNU-x86_64d\target\3.0\qt5_15-GNU-x86_64d	PATH NOT FOUND	
- QueryOpen	plugins\gammaray\3.0\qt5_15-GNU-x86_64d\target	NAME NOT FOUND	
- QueryOpen	bin\gammaray\3.0\qt5_15-GNU-x86_64d\target	PATH NOT FOUND	
- QueryOpen	bin\gammaray-target	NAME NOT FOUND

After that, my application is started, bot no GammaRay window is shown. 

[...]

PR #1082, opened 2025-08-02
Fix GammaRay GUI not showing up on Windows

iamsergio:
The target sends the TCP address via a QLocalSocket, but Windows is very prone to partial writes.

Ensure all bytes are written before closing the socket.

Simplified the launcher code, no need to use a while, as the target only sends 1 message. Ignore the readyRead that contain partial writes.

Related to some comments in issue #959

### DIFF
--- a/core/probesettings.cpp
+++ b/core/probesettings.cpp
@@ -139,7 +139,7 @@ void ProbeSettingsReceiver::readyRead()
     }
 }
 
-inline void waitForBytesWritten(QLocalSocket &socket)
+static void waitForBytesWritten(QLocalSocket &socket)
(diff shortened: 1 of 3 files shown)


core/probesettings.cpp
```diff
@@ -139,6 +139,16 @@ void ProbeSettingsReceiver::readyRead()
     }
 }
 
+inline void waitForBytesWritten(QLocalSocket &socket)
```
Waqar144, line 142: should be marked `static`?

Waqar144 (approved):
Looks good! Thanks for fixing this!

PR #1125, opened 2026-04-15
Fix attach crash on complex Qt applications

Kevinye0116:
## Problem

When attaching GammaRay to complex Qt applications (e.g., apps using
Qt Quick, QQuickWindow, QSettings, etc.), the target process crashes
with an access violation in `Qt6Core!QSettings::d_func` (null `this`
pointer). Simple applications were unaffected, and Launch mode worked
correctly in all cases.

Root cause: `gammaray_probe_attach()` was not calling
`Hooks::installHooks()` or `Probe::startupHookReceived()`, leaving
the probe without real-time object lifecycle tracking during
`findExistingObjects()` traversal. This caused access to freed/invalid
objects in multi-threaded complex applications.

### probe/hooks.cpp
- Added `Hooks::installHooks()` call in `gammaray_probe_attach()` to
  register Qt object creation/destruction hooks, consistent with the
  Launch mode flow.
- Added `Probe::startupHookReceived()` call to set `trackDestroyed =
  false`, preventing unsafe destruction tracking during initial object
  discovery.

### core/probe.cpp
- Added null checks for child objects in `Probe::discoverObject()`.
- Added `m_validObjects.contains(child)` check to skip already-
  discovered objects and prevent redundant recursive traversal.
- Replaced `foreach` with range-based for loop.

### plugins/quickinspector/quickinspector.cpp
- Added null checks for `view->engine()`, `window->contentItem()`,
  and `children.first()` in `QuickInspector::objectCreated()` to
  guard against partially-initialized QQuickWindow states during
  attach-time object discovery.

## Testing

| Scenario                        | Before | After  |
| ------------------------------- | ------ | ------ |
| Attach + complex Qt application | Crash  | ✅ OK  |
| Attach + simple application     | OK     | ✅ OK  |
| Launch + complex application    | OK     | ✅ OK  |
| Launch + simple application     | OK     | ✅ OK  |

## Known Remaining Issues

- **Events view**: No upper bound on stored event count; may cause
  unresponsiveness over time with high-throughput applications.
- **Objects view**: First load may be slow for applications with a
  large number of top-level QObjects.

These are tracked but not addressed in this PR.


sergio-kdab:
can we get a test that reproduces the problem ?

Kevinye0116:
> can we get a test that reproduces the problem ?

I tried attaching GammaRay to Insta360 Studio, and it crashed at first. The issue has been resolved after the modifications. The Quick Scenes interface experiences noticeable lag when loading some QML components, and the issue is still unresolved.

Kevinye0116:
> [CLA assistant check](https://cla-assistant.io/KDAB/GammaRay?pullRequest=1125) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/KDAB/GammaRay?pullRequest=1125) before we can accept your contribution.**1** out of **2** committers have signed the CLA.✅ Kevinye0116❌ pre-commit-ci[bot]You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/KDAB/GammaRay?pullRequest=1125) it.

I have already signed the CLA.
It looks like the remaining committer is pre-commit-ci[bot].
Could you please recheck or exclude the bot from CLA validation?

